PR #5, opened 2022-10-07
Fix error in install script

dpo:
This causes a failure on macOS. The fix should also work on Linux.


nimgould:
Sorry, I missed this. Fixed on main branch

PR #10, opened 2022-10-09
GH Action to build and test

dpo:


dpo:
@nimgould Here is a GitHub Action to automatically build a standalone GALAHAD for Linux and macOS. By clicking on the "Details" link below 👇   ("standalone / ubuntu" or "standalone / macos"), you can see the build steps and the details of the build. Currently, both fail because MA27 is not present.

If we merge this pull request and you submit your changes in a pull request, this action will run automatically and you'll be able to assess directly whether the builds succeed.

ps: the fact that the installation process is interactive is a real obstacle here. (It's the same in CUTEst & Co.) We should really think of modernizing the install script (at the very least). The process is brittle because the "user selections" change each time you add a platform, a compiler, or some other interactive option.

dpo:
@nimgould A few updates here:
1. I managed to build GALAHAD on Linux by installing `libhwloc-dev` with `apt-get`. I haven't gotten Homebrew hwloc to work (I think because of a glibc version mismatch but need to investigate more). This will be an issue for Homebrew users on Linux. However, there are issues in the unit tests:
    * In the C interface unit tests, there are some statuses = 18 towards the end. Is that a problem? I was expecting status = 0;
    * LANCELOT-B segfaults;
    * Tests related to the SIF interface for LANCELOT shouldn't be run (I did not install the SIF interface).

2. GALAHAD still doesn't install on macOS because of the SSIDS issue.

nimgould:
1. I would expect some nonzero status values if the requested linear solvers are not present. 
    I have modified both sls and ils to try to detect this, but this is ongoing.  Need to watch, and
    also to test with HSL present to see if these disappear. 
2. Need to find out why. I will check here.
3. Agreed. Need to check for this. On my list

dpo:
@nimgould I update the CI to the latest ARCHDefs and GALAHAD. We see the shared lib for macOS fail to build. I'll have a look when I can.

dpo:
@nimgould I refreshed this with the latest changes in `master`. There are compilation issues on Linux starting here:
https://github.com/ralna/GALAHAD/actions/runs/3412660790/jobs/5678455681#step:7:393

dpo:
Hi @nimgould. I rebased this on top of the latest changes in `master`. On Linux, the unit tests are still trying to run the SIF interface to LANCELOT:
```
 Testing SIF interface to LANCELOT B
/bin/sh: 1: sdgal: not found
make[1]: *** [/home/runner/work/GALAHAD/GALAHAD/src/lancelot/makemaster:208: test_cutest_single] Error 127
make: *** [/home/runner/work/GALAHAD/GALAHAD/src/makemaster:421: tests_single] Error 2
Error: Process completed with exit code 2.
```

On macOS, still lots of undefined symbols when building the shared libs: https://github.com/ralna/GALAHAD/actions/runs/3439387926/jobs/5736620985#step:7:710

I can probably help with that. Has the shared library building process stabilized on Linux?

nimgould:
I believe that this is because you are still using the makefile stanza "tests" not "test" in the workfile. "test" just performs the comprehensive stand-alone tests while "tests" does this + the sif/cutest tests.

Yes, please, for the shared libs. I think you need to check with @amontoison

amontoison:
For the shared libraries, I need your help @nimgould.
For the Julia interface, I checked what we need and the conclusion is we only need the shared library `libgalahad_c`.

We can't use the "glue" workaround here and also this hack is only working on Linux for information (even if you just want the Python interface).
The last issue that I have is https://github.com/ralna/GALAHAD/issues/19 and can't continue without a fix.

Nick, if we can't compile, `-Wl,--undefined`, It can't work on Windows or Mac.
Windows doesn't allow incomplete shared libraries and we need to tune the option of Mac compilers to allow undefined symbols.

dpo:
@nimgould Thank you, I have changed "tests" to "test" now.

A new round of tests with fresh code. I also added continuous integration with icc/ifort on Linux and macOS.

Here's the rundown:
* macOS/Intel: error when creating the shared libs that looks like a bug in variable substitution (https://github.com/ralna/GALAHAD/actions/runs/3449548354/jobs/5757557343#step:9:342)
* Linux/Intel: SLS seems to be looking for MC61 (https://github.com/ralna/GALAHAD/actions/runs/3449548354/jobs/5757557303#step:11:123)
* macOS/GNU: same error with undefined symbols when creating the shared libs (https://github.com/ralna/GALAHAD/actions/runs/3449548354/jobs/5757557273#step:9:704)
* Linux/GNU: SLS seems to be looking for MC77 (https://github.com/ralna/GALAHAD/actions/runs/3449548354/jobs/5757557247#step:11:97)

Interestingly, the errors with GNU and Intel compilers are different.

nimgould:
1. & 3 I can only presume that one of your variables CC, FORTRAN, OPTIMIZATION or SHARED has more than one term. The script used positional parameters. I have changed this so that instead it picks up shell variables before the call.

2 & 4 Yes, the sed translate missed these two (and mc21bd). single.sed updated.

Well spotted. Thanks!

dpo:
Rebased again. Here's the situation:

* macOS/Intel: there's an inexistent hardcoded path in ARCHDefs that causes an error (https://github.com/ralna/GALAHAD/actions/runs/3451480609/jobs/5760728211#step:9:339). I'm removing it now but having this error locally that is related to SSIDS:
```
In file included from ./ssids/cpu/factor.hxx(23),
                 from ssids/cpu/NumericSubtree.hxx(10),
                 from NumericSubtree.cxx(6):
./ssids/cpu/Workspace.hxx(37): error: more than one instance of overloaded function "std::align" matches the argument list:
            function "std::__1::align(size_t={__darwin_size_t={unsigned long}}, size_t={__darwin_size_t={unsigned long}}, void *&, size_t={__darwin_size_t={unsigned long}} &)" (declared at line 839 of "/Library/Developer/CommandLineTools/usr/bin/../../usr/include/c++/v1/memory")
            function "std::align(size_t={__darwin_size_t={unsigned long}}, size_t={__darwin_size_t={unsigned long}}, void *&, size_t={__darwin_size_t={unsigned long}} &)" (declared at line 20 of "./compat.hxx")
            argument types are: (const int, size_t, void *, size_t)
        if(!std::align(align, sz, mem_aligned_, sz_)) throw std::bad_alloc();
            ^

In file included from ./ssids/cpu/factor.hxx(24),
                 from ssids/cpu/NumericSubtree.hxx(10),
                 from NumericSubtree.cxx(6):
./ssids/cpu/kernels/assemble.hxx(207): warning #3180: unrecognized OpenMP #pragma
        #pragma omp taskgroup
        ^
```
* Linux/Intel: undefined reference to `spral::ssids::cpu::host_gemm<double>()` (https://github.com/ralna/GALAHAD/actions/runs/3451480609/jobs/5760728182#step:11:112)
* macOS/GNU: same as above;
* Linux/GNU: same undefined symbol in SSIDS (https://github.com/ralna/GALAHAD/actions/runs/3451480609/jobs/5760728086#step:11:98)

nimgould:
For the first one, I need more details, so please, the call to create_shared at the end of

/usr/bin/make -f /Users/runner/work/GALAHAD/GALAHAD/makefiles/mac64.osx.ifr create_shared

I also need to see the $GALAHAD/makefiles/mac64.osx.ifr to see what settings you have.

For the others, can you go into src/eqp and then

/usr/bin/make -f /Users/runner/work/GALAHAD/GALAHAD/makefiles/mac64.osx.ifr eqpdt

I am presuming that you simply have a missing standard library

dpo:
I did not change the default. Here is my `makefiles/mac64.osx.ifr`: https://gist.github.com/436d13f4eb9b3724c9fc00129d93aaff

I ran the command `make -f ../makefiles/mac64.osx.ifr create_shared` from `$GALAHAD/src`. Is that where I was supposed to run it from? It resulted in lots of undefined symbols (similar to the macOS/GNU build). However, I don't think the error I pasted above occurred while making shared librairies. It occurred while compiling `NumericSubtree`.

Investigating the other failures on the VMs takes more time.

nimgould:
I can't see any failure when compiling  NumericSubtree in the above. All the failures were in building the shared library. Am I looking in the wrong place?

Thanks for the mac64.osx.ifr file. The only strange token is RANLIB = true

dpo:
You're not looking in the wrong place. The error I quoted above occurs locally on my Mac. The differences with the GitHub virtual machines are:
- I have a Silicon (aarch64) processor while the VM uses and Intel architecture;
- I'm using the Intel compilers version 2022.2.1 while the VM uses 2022.3.1.

I'l rerun the builds with the updated ARCHDefs.

There used to be no ranlib on macOS; running it would only output annoying warning messages. It seems like the behavior has changed on my machine though. I'll look into it but I don't think the error above is related to ranlib.

dpo:
Here's the error with the updated ARCHDefs: https://github.com/ralna/GALAHAD/actions/runs/3456046277/jobs/5768502085#step:9:185

nimgould:
It seems from

error: more than one instance of overloaded function "std::align" 
matches the argument list:

that there is a choice of possible std::align, but that these are
not compatible. So much for standard!

Can you go into src/ssids and try

  make -f $GALAHAD/makefiles/mac64.osx.ifr NumericSubtree.o

This will at least tell us what flags are being passed

Thank you
-- 
Nicholas I. M. Gould
Scientific Computing Department            email: ***@***.***
G39, R71 Rutherford Appleton Laboratory    phone: 0(44)1235 445801
Chilton, Oxon OX11 0QX England EU
www: http://www.numerical.rl.ac.uk/people/nimg

This email and any attachments are intended solely for the use of the named recipients. If you are not the intended recipient you must not use, disclose, copy or distribute this email or any of its attachments and should notify the sender immediately and delete this email from your system. UK Research and Innovation (UKRI) has taken every reasonable precaution to minimise risk of this email or any attachments containing viruses or malware but the recipient should carry out its own virus and malware checks before opening the attachments. UKRI does not accept any liability for any losses or damages which the recipient may sustain due to presence of any viruses.

dpo:
@nimgould Here is the log: https://gist.github.com/d056ea7a421d090efe30a71319f1cf7a

nimgould:
Sounds reasonable, Jari I suspect you need to test this in SPRAL before
we pass it into GALAHAD.

-- 
Nicholas I. M. Gould
Scientific Computing Department            email: ***@***.***
G39, R71 Rutherford Appleton Laboratory    phone: 0(44)1235 445801
Chilton, Oxon OX11 0QX England EU
www: http://www.numerical.rl.ac.uk/people/nimg

This email and any attachments are intended solely for the use of the named recipients. If you are not the intended recipient you must not use, disclose, copy or distribute this email or any of its attachments and should notify the sender immediately and delete this email from your system. UK Research and Innovation (UKRI) has taken every reasonable precaution to minimise risk of this email or any attachments containing viruses or malware but the recipient should carry out its own virus and malware checks before opening the attachments. UKRI does not accept any liability for any losses or damages which the recipient may sustain due to presence of any viruses.

jfowkes:
I've created a SPRAL pull request here:
https://github.com/ralna/spral/pull/100
@nimgould all tests pass, could you test with GALAHAD and if everything works approve?

jfowkes:
@dpo to make homebrew work on linux you now need to do:
`eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"`
see the [ubuntu-latest docs](https://github.com/actions/runner-images/blob/releases/ubuntu20/20221027/images/linux/Ubuntu2004-Readme.md) for more info.

nimgould:
Will do, but before I do, can @dpo test with the updated compat.hxx
that I just pushed. Works fine here with both linux compilers and gcc


This email and any attachments are intended solely for the use of the named recipients. If you are not the intended recipient you must not use, disclose, copy or distribute this email or any of its attachments and should notify the sender immediately and delete this email from your system. UK Research and Innovation (UKRI) has taken every reasonable precaution to minimise risk of this email or any attachments containing viruses or malware but the recipient should carry out its own virus and malware checks before opening the attachments. UKRI does not accept any liability for any losses or damages which the recipient may sustain due to presence of any viruses.

dpo:
Thanks both. That issue seems to be resolved. The next error is
```
./ssids/cpu/kernels/assemble.hxx(231): error: no instance of constructor "std::__1::vector<_Tp, _Allocator>::vector [with _Tp=int, _Allocator=std::__1::__allocator_traits_rebind_t<spral::ssids::cpu::BuddyAllocator<<unnamed>::T={double}, std::__1::allocator<<unnamed>::T={double}>>, int>]" matches the argument list
            argument types are: (int, PoolAllocInt)
     std::vector<int, PoolAllocInt> map(n+1, PoolAllocInt(pool_alloc));
                                        ^
```
and I'm getting a lot of these warnings:
```
In file included from NumericSubtree.cxx(13):
ssids/cpu/AppendAlloc.hxx(85): warning #3180: unrecognized OpenMP #pragma
        #pragma omp critical
        ^
```
Complete log: https://gist.github.com/ce0e76f84d48fc8b2d1314c4c9538e88

@jfowkes I notice that the SPRAL CI only tests Linux and gcc. You would catch those errors with an expanded CI that includes macOS and the Intel compilers, as I did here.

dpo:
@jfowkes

> to make homebrew work on linux you now need to do:
`eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"`
see the [ubuntu-latest docs](https://github.com/actions/runner-images/blob/releases/ubuntu20/20221027/images/linux/Ubuntu2004-Readme.md) for more info.

Thanks. I think that's already what I'm doing [here](https://github.com/ralna/GALAHAD/pull/10/files#diff-92f23bd195b71e42d05a8b81008449b087d57dc0accd3916d94dcfb5abadaabcR57), isn't it? The Homebrew action should put `brew` on the path (and indeed, on Linux, it's under `/home/linuxbrew/.linuxbrew/bin`).

nimgould:
I have no idea what "no instance of constructor "std::__1::vector"  means. I gather from google that std::__1::vector is a standard rename of std::vector from the g++ library, but after that it is all c++ which I don't follow. The ./ssids/cpu/kernels/assemble.hxx file seems to define PoolAlloc type, and from what I understand std::vector is supposed to map an index (n+1 in this case) to an instance of this type ... but then it complains that it can't reconcile PoolAlloc with anything it knows about.

@jfowkes what happens if you test spral on your Mac? Do you see this issue?

jfowkes:
@nimgould sorry that didn't completely remove the `std::align` macro, could you test and merge the following:
https://github.com/ralna/spral/pull/101
This also removes the defining code and autoconf stuff that looks for it (basically just cleans up the codebase).

nimgould:
I was about to suggest this, and it removes many of the messages; I now use
CXXBASIC  = -c -std=c++11 -fpic -I/usr/local/include -fopenmp -diag-disable=10441
in the GALAHAD makemaster (perhaps some kind soul could add this to ccompiler.pc*.lnx.ic*"
in archdefs ... I haven't started to master the git black arts at the moment, and so can't make a pull request :).

We do get a bunch of 
ldlt_app.cxx(1550): warning #3850: cancellation point with no cancel in region
                      #pragma omp cancellation point taskgroup
messages, don't know what this means or how to fix it, but thankfully it is only a warning

dpo:
I made corresponding changes to `ccompiler.mac64.osx.icc`. The OpenMP warnings are gone. I ran a fresh set of builds and here's where we are:

1. macOS/GNU: strangely, `hwloc.h` can't be found. This is a regression because it used to be found. I tried setting `HWLOC_CFLAGS` and `HWLOC_LIBS` as in SPRAL, but that had no effect. It seems building SSIDS here and in SPRAL works differently.
2. linux/GNU: undefined reference to `void spral::ssids::cpu::host_gemm` in the tests.
3. macOS/Intel: the `std::vector` allocator error remains. I guess we have to compile with `std=c++14` here.
4. linux/Intel: `undefined reference to __kmpc_global_thread_num` and many others.

Details in the checks below 👇.

jfowkes:
Thanks @dpo the `std::vector` allocator error is fixed with either `-std=c++14` or `-std=c++17`.

I agree that we should default to OpenBLAS as the default BLAS/LAPACK, it is excellent and freely available everywhere. On macOS the Accelerate framework still ships with a 13 year old (yes really) buggy LAPACK version.

nimgould:
For 4, I believe that setting OPENMP = - fopenmp sorts things out. Thanks to @jfowkes for that.
For 1

dpo:
What about number 3? Should we add `-std=c++14` (or 17) to the ARCHDefs definitions? Why isn't that flag needed on LInux?

Also, does GALAHAD run SPRAL's `configure` script? How can I apply the flags that I determined [there](https://github.com/ralna/spral/pull/97#issuecomment-1315967393) here?

About number 2: how do I tell GALAHAD what BLAS/LAPACK I want to use? The documentation says GALAHAD provides a default BLAS, but I guess it doesn't?!

amontoison:
BLAS / LAPACK routines are in the folder `src/lapack`:
https://github.com/ralna/GALAHAD/tree/master/src/lapack

nimgould:
For 3, it doesn't seem to be necessary for Linux, code compiles and runs fine without. If it is needed for macos, then, yes, make the change.

No, GALAHAD doesn't use SPRAL's configure script, the archdefs files were constructed by deciphering what configure did, and making sure it didn't conflict with other galahad needs.

There are default libraries LAPACK = galahad_lapack and BLAS = -lgalahad_blas that are compiled at install time if the values 
LAPACk=
BLAS=
are set in the appropriate archdef file. We can override these so long as we are confident that there are better tuned versions available;  installers are warned in the README to change these if there are machine-specific versions available when given the opportunity by the install script. So if you feel for Macos that there will always be better versions available, by all means set them. Of course, how this then plays with your demands that Macos dylib files are complete, I have no idea ... indeed, I really don't know how OSX copes with mixing libraries (Linux does this at load time when missing function calls are resolved by searching through a list of available libraries)

dpo:
> For 3, it doesn't seem to be necessary for Linux, code compiles and runs fine without. If it is needed for macos, then, yes, make the change.

I don't understand how that could be and I find it a little suspicious. The C++ standard isn't different on Linux and macOS. On Linux, we build with `-std=c++11`. How could that be? Compiler bug?

About BLAS/LAPACK, sorry for being thick, but this looks to me as a bug in the build system. The error occurs for Linux with gfo/gcc in the unit tests and is related to SSIDS: https://github.com/ralna/GALAHAD/actions/runs/3475054150/jobs/5808868839#step:11:98
```
NumericSubtree.cxx:(.text._ZN5spral5ssids3cpu17factor_node_indefIdNS1_14BuddyAllocatorIdSaIdEEEEEviRKNS1_12SymbolicNodeERNS1_11NumericNodeIT_T0_EERKNS1_18cpu_factor_optionsERNS1_11ThreadStatsERSt6vectorINS1_9WorkspaceESaISK_EERSB_[_ZN5spral5ssids3cpu17factor_node_indefIdNS1_14BuddyAllocatorIdSaIdEEEEEviRKNS1_12SymbolicNodeERNS1_11NumericNodeIT_T0_EERKNS1_18cpu_factor_optionsERNS1_11ThreadStatsERSt6vectorINS1_9WorkspaceESaISK_EERSB_]+0x3a4): undefined reference to `void spral::ssids::cpu::host_gemm<double>(spral::ssids::cpu::operation, spral::ssids::cpu::operation, int, int, int, double, double const*, int, double const*, int, double, double*, int)'
```
The ARCHDefs files leave `BLAS` and `LAPACK` empty. If I understand well, that means `libgalahad_blas` and `libgalahad_lapack` are linked in. So how could `gemm` be missing?

We have to keep in mind that when a build error occurs on the pristine virtual machine and not on our own, it's likely because of something specific we did on our own machines. And users will likely face the same issues.

jfowkes:
> > For 3, it doesn't seem to be necessary for Linux, code compiles and runs fine without. If it is needed for macos, then, yes, make the change.
> 
> I don't understand how that could be and I find it a little suspicious. The C++ standard isn't different on Linux and macOS. On Linux, we build with `-std=c++11`. How could that be? Compiler bug?

In the SPRAL CI tests we only need a higher C++ standard for Clang, so either there's a bug in Clang or GCC/Intel implemented that particular feature early before it got standardised.

nimgould:
I have no evidence for the former, except that it works on Linux. Of course it won't hurt to upgrade to c++14 for consistency. Johnathan Hogg probably wasn't using c++17, so if 14 works,
that would be preferable (I have no idea if C evolves in an upwardly compatible way like fortran).

Can we see what is happening in the eqp failure. go into the ./src/eqp directory and

make -f $GALAHAD/makefiles/pc64.lnx.gfo eqpdt

and see what that links in. I should add that ssids was only ever intended to be in double precision, and I notice that in the report

/usr/bin/ld: /home/runner/work/GALAHAD/GALAHAD/objects/pc64.lnx.gfo/single/libgalahad_spral.a(NumericSubtree.o): in function `void spral::ssids::cpu::factor_node_indef<double, spral::ssids::cpu::BuddyAllocator<double, std::allocator<double> > >(int, spral::ssids::cpu::SymbolicNode const&, spral::ssids::cpu::NumericNode<double, spral::ssids::cpu::BuddyAllocator<double, std::allocator<double> > >&, spral::ssids::cpu::cpu_factor_options const&, spral::ssids::cpu::ThreadStats&, std::vector<spral::ssids::cpu::Workspace, std::allocator<spral::ssids::cpu::Workspace> >&, spral::ssids::cpu::BuddyAllocator<double, std::allocator<double> >&)':

there are an awful lot of mentions of the word double!

dpo:
> I should add that ssids was only ever intended to be in double precision

Oh boy.

nimgould:
Indeed.  @jfowkes do you know if anyone has tried spral in single precision?

jfowkes:
> Indeed. @jfowkes do you know if anyone has tried spral in single precision?

I don't think so, as far as I know all the test are in double precision.

I'm starting to think that relying on SSIDS is a bad idea...

nimgould:
We can always drop through to sytrf/s if ssids gives us errors in single precision. Not ideal, but then single precision optimization rarely is (unless you are Roger Fletcher, who believed that all algorithms should be implemented in low precision as a check for robustness!)

nimgould:
I think its fine if we use it as it was designed (in double).

dpo:
Oh I disagree. I believe there's a lot of potential in single precision (and lower precisions) for the future of computing.

So isn't it easier to simply disable SSIDS if the user is building in single precision? When we have MUMPS, we can make it the default.

nimgould:
IT probably does mean that I'll have to add some extra seds to the c++ parts to convert doubles to floats and the like. The fortran parts of ssids already have this enabled.

I do agree that there is potential for low precision, but it is a shame that (at least) intel chips
currently simply fill half the double-precision registers with zeros and then use double. But the ARM chips should change all of this.

dpo:
It sounds like you could use an intern to convert SSIDS to single precision.

jfowkes:
We would love that, but it requires a deep understanding of Fortran, C/C++ and CUDA.

dpo:
This is an unfortunate example of high technical debt. Code that nobody understands or can even modify, but that we need to use anyways.

amontoison:
What version of METIS is required by `SPRAL` ? Is it the version 4 or the version 5?
If it's the version 5, we will have some issues with HSL packages because they need the version 4.

dpo:
I built SPRAL with METIS 5 locally and it seems to be working. In `COMPILE.md`, they recommend to install METIS from the COIN-OR project, which downloads METIS 4: https://github.com/coin-or-tools/ThirdParty-Metis/blob/stable/2.0/get.Metis#L33

amontoison:
I asked this question because two shared libraries libmetis.$dlext can't coexist.

dpo:
Ok, so the CI is now at a point where
* both linux versions fail in the unit tests when using SSIDS with single precision GALAHAD;
* both macOS versions fail when building the dynamic libraries.

My next task is to look into those dynamic libraries.

nimgould:
SPRAL can take either Metis. HSL needs metis 4, MUMPS and Pastix need metis 5. I presume that we can simply have a libmetis4.a/so and libmetis.a/so (for v5). I believe that the functions in 4 and 5 can coexist as they have different signatures. I shall rename the dummy metis4 in galahad to be galahad_metis4.a to make this clear.

nimgould:
More broadly, I now think we should provide two variables METIS4 and METIS5 in every compiler.* file in ARCHDEFS, setting as defaults

 METIS4='-lgalahad_metis4'
 METIS5=

This will then permeate into the $GALAHAD/makefiles/* on install, and will provide users opportunities subsequently to replace these with real versions of metis4 and metis5 if they are available. Incidentally, metis5 (and parametis 4) is available as part of the Linux distributions, as is mumps with or without (p)scotch orderings. Is that true on Macs as well? I have always downloaded and compiled metis 4 as described in the README.external file in galahad. 

What do you think @dpo @jfowkes @amontoison ? If you agree, someone will have to make a global change to the archdefs files, and we need to coordinate the change with those for galahad so that they stay in sync

nimgould:
I have turned off the build of ssids in the single precision case for the time being. I had a look at the c++ an cuda parts that would need to be auto-sed-matically  translated, but I do not feel confident enough in my C (or at all in my C++) to be sure that I got this right. Maybe one day.

jfowkes:
Happy to have a look at this with you at some point, but no promises we'll be able to make it work.

nimgould:
I am afraid that ssids in double precision is as leaky as a paper bag in a storm.  Even on the small comprehensive test program, valgrind reports 

=4076300== LEAK SUMMARY:
==4076300==    definitely lost: 72 bytes in 3 blocks
==4076300==    indirectly lost: 10,357,512 bytes in 43 blocks
==4076300==      possibly lost: 16,777,248 bytes in 2 blocks
==4076300==    still reachable: 456 bytes in 3 blocks
==4076300==         suppressed: 0 bytes in 0 blocks

all of which relate to ssids c++ creates. Not good

nimgould:
Thank you @jfowkes , but not high priority I fear

jfowkes:
Yes unfortunately we have several open bugs in SPRAL relating to memory leaks, but no idea how to fix them...

dpo:
> Incidentally, metis5 (and parametis 4) is available as part of the Linux distributions, as is mumps with or without (p)scotch orderings. Is that true on Macs as well?

None of that is available on Macs by default in any official way. Precompiled binaries are usually available from Homebrew. METIS 5 is available but more obscure packages like MUMPS and ParMETIS are not officially available (I maintain the "taps" in my "free" time).

I would be very reluctant to rely on Linux package managers as they are distribution dependent. Unless you're on Ubuntu 22, you're stuck with buggy MUMPS 5.2 from 5 years ago. What about the 2000 other linux distros? Though it's not perfect, I've found that Homebrew is my preferred package manager on linux too.

amontoison:
@dpo, I think we should provide precompiled binaries of GALAHAD with Yggdrasil even if the user is not using Julia.
It will be probably enough for 95% of the users.

amontoison:
@nimgould @jfowkes 
We forgot to share with you that precompiled binaries are available online here:
https://github.com/JuliaBinaryWrappers/SPRAL_jll.jl
https://github.com/JuliaBinaryWrappers/SPRAL_jll.jl/releases/tag/SPRAL-v0.1.0%2B0

It means that a user can just download an archive of SPRAL (and its dependency) for its platform and it works.

dpo:
Here's the latest on the shared libraries for macOS. Firsly, I have to ask gfortran to compile with -O2 in order to work around a bug in the linker that should be fixed in the next release of the command-line tools. I'll open a pull request to ARCHDefs.

Secondly, if I install GALAHAD without the shared libraries, and then call `create_one_shared` (from which I removed `--no-undefined`), I get the following undefined symbols
```
❯ make -f /Users/dpo/dev/ralna/GALAHAD/makefiles/mac64.osx.gfo create_one_shared                        13:15:16
cd /Users/dpo/dev/ralna/GALAHAD/objects/mac64.osx.gfo/double; CC=gcc-12 FORTRAN=gfortran-12 OPTIMIZATION=-O2 \
                   SHARED=-shared DLEXT=dylib LOADALL=-all_load \
                   LOADNONE=-noall_load \
                   sh /Users/dpo/dev/ralna/GALAHAD/bin/create_one_shared
 creating single GALAHAD shared library in
  /Users/dpo/dev/ralna/GALAHAD/objects/mac64.osx.gfo/double/shared
 creating libgalahad_all.dylib
gfortran-12 -O2 -shared -o shared/libgalahad_all.dylib -Wl,-all_load libgalahad.a libgalahad_hsl.a libgalahad_spral.a             libgalahad_mkl_pardiso.a libgalahad_pardiso.a libgalahad_wsmp.a             libgalahad_pastix.a libgalahad_mumps.a libgalahad_umfpack.a             libgalahad_metis.a libgalahad_lapack.a libgalahad_blas.a             libgalahad_cutest_dummy.a libgalahad_hsl_c.a libgalahad_c.a -Wl,-noall_load -lstdc++ -lgomp
ld: warning: option -noall_load is obsolete and being ignored
Undefined symbols for architecture arm64:
  "___hsl_ma77_double_MOD_ma77_alter_double", referenced from:
      ___galahad_sls_double_MOD_sls_alter_d in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_analyse_double", referenced from:
      ___galahad_sls_double_MOD_sls_analyse in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_enquire_indef_double", referenced from:
      ___galahad_sls_double_MOD_sls_enquire in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_enquire_posdef_double", referenced from:
      ___galahad_sls_double_MOD_sls_enquire in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_factor_double", referenced from:
      ___galahad_sls_double_MOD_sls_factorize in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_finalise_double", referenced from:
      ___galahad_sls_double_MOD_sls_terminate in libgalahad.a(sls.o)
      ___galahad_sls_double_MOD_sls_factorize in libgalahad.a(sls.o)
      ___galahad_sls_double_MOD_sls_analyse in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_input_reals_double", referenced from:
      ___galahad_sls_double_MOD_sls_factorize in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_input_vars_double", referenced from:
      ___galahad_sls_double_MOD_sls_analyse in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_open_double", referenced from:
      ___galahad_sls_double_MOD_sls_analyse in libgalahad.a(sls.o)
      ___hsl_ma77_double_iface_MOD_ma77_open_main in libgalahad_hsl_c.a(hsl_ma77d_ciface.o)
  "___hsl_ma77_double_MOD_ma77_restart_double", referenced from:
      ___galahad_sls_double_MOD_sls_factorize in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_scale_double", referenced from:
      ___galahad_sls_double_MOD_sls_factorize in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_solve_double", referenced from:
      ___galahad_sls_double_MOD_sls_solve_multiple_rhs.constprop.0 in libgalahad.a(sls.o)
      ___galahad_sls_double_MOD_sls_solve_one_rhs.constprop.0 in libgalahad.a(sls.o)
      ___galahad_sls_double_MOD_sls_sparse_forward_solve in libgalahad.a(sls.o)
      ___galahad_sls_double_MOD_sls_part_solve in libgalahad.a(sls.o)
  "___hsl_ma77_double_MOD_ma77_solve_fredholm_double", referenced from:
      ___galahad_sls_double_MOD_sls_fredholm_alternative in libgalahad.a(sls.o)
  "___hsl_ma86_double_MOD_ma86_analyse_double", referenced from:
      ___galahad_sls_double_MOD_sls_analyse in libgalahad.a(sls.o)
  "___hsl_ma86_double_MOD_ma86_factor_double", referenced from:
      ___galahad_sls_double_MOD_sls_factorize in libgalahad.a(sls.o)
  "___hsl_ma86_double_MOD_ma86_finalise_double", referenced from:
      ___galahad_sls_double_MOD_sls_terminate in libgalahad.a(sls.o)
  "___hsl_ma86_double_MOD_ma86_solve_mult_double", referenced from:
      ___galahad_sls_double_MOD_sls_solve_multiple_rhs.constprop.0 in libgalahad.a(sls.o)
  "___hsl_ma86_double_MOD_ma86_solve_one_double", referenced from:
      ___galahad_sls_double_MOD_sls_solve_one_rhs.constprop.0 in libgalahad.a(sls.o)
      ___galahad_sls_double_MOD_sls_sparse_forward_solve in libgalahad.a(sls.o)
      ___galahad_sls_double_MOD_sls_part_solve in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_alter_double", referenced from:
      ___galahad_sls_double_MOD_sls_alter_d in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_analyse_double", referenced from:
      ___galahad_sls_double_MOD_sls_analyse in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_enquire_indef_double", referenced from:
      ___galahad_sls_double_MOD_sls_enquire in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_enquire_posdef_double", referenced from:
      ___galahad_sls_double_MOD_sls_enquire in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_factor_double", referenced from:
      ___galahad_sls_double_MOD_sls_factorize in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_finalise_double", referenced from:
      ___galahad_sls_double_MOD_sls_terminate in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_solve_double", referenced from:
      ___galahad_sls_double_MOD_sls_solve_multiple_rhs.constprop.0 in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_solve_fredholm_double", referenced from:
      ___galahad_sls_double_MOD_sls_fredholm_alternative in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_solve_one_double", referenced from:
      ___galahad_sls_double_MOD_sls_solve_one_rhs.constprop.0 in libgalahad.a(sls.o)
      ___galahad_sls_double_MOD_sls_part_solve in libgalahad.a(sls.o)
  "___hsl_ma97_double_MOD_ma97_sparse_fwd_solve_double", referenced from:
      ___galahad_sls_double_MOD_sls_sparse_forward_solve in libgalahad.a(sls.o)
  "_hwloc_get_nbobjs_by_depth", referenced from:
      __ZNK5spral11hw_topology13HwlocTopology14get_numa_nodesEv in libgalahad_spral.a(guess_topology.o)
  "_hwloc_get_obj_by_depth", referenced from:
      __ZNK5spral11hw_topology13HwlocTopology14get_numa_nodesEv in libgalahad_spral.a(guess_topology.o)
  "_hwloc_get_type_depth", referenced from:
      __ZNK5spral11hw_topology13HwlocTopology14get_numa_nodesEv in libgalahad_spral.a(guess_topology.o)
  "_hwloc_topology_destroy", referenced from:
      _spral_hw_topology_guess in libgalahad_spral.a(guess_topology.o)
  "_hwloc_topology_init", referenced from:
      _spral_hw_topology_guess in libgalahad_spral.a(guess_topology.o)
  "_hwloc_topology_load", referenced from:
      _spral_hw_topology_guess in libgalahad_spral.a(guess_topology.o)
  "_hwloc_topology_set_type_filter", referenced from:
      _spral_hw_topology_guess in libgalahad_spral.a(guess_topology.o)
ld: symbol(s) not found for architecture arm64
collect2: error: ld returned 1 exit status
make: *** [create_one_shared] Error 1
```
I can resolve the `hwloc` symbols by adding `-L/opt/homebrew/opt/hwloc/lib -lhwloc` to `create_one_shared`, but the missing HSL symbols remain.

amontoison:
Thanks to `Clang.jl`, I explicitly listed the dependency of each GALAHAD packages with a C interface [here](https://github.com/amontoison/GALAHAD.jl/blob/main/src/GALAHAD.jl#L5-L143) if it can help.
HSL packages were used almost everywhere!

nimgould:
I really don't understand what the mac loader is doing. All of the dependencies should be in libgalahad_hsl.a. Could you try an ar t on libgalahad_hsl.a (in your/Users/dpo/dev/ralna/GALAHAD/objects/mac64.osx.gfo/double). Here I get 

% ar t libgalahad_hsl.a 
dummy_hsl.o
ma61d.o
ma27d.o
mc71d.o
ma57d.o
mc21d.o
mc34d.o
mc47d.o
mc59d.o
mc64d.o
hsl_ma57d.o
hsl_ad02d.o
kb07i.o
hsl_kb22l.o
hsl_ma54d.o
hsl_ma64d.o
hsl_of01i.o
hsl_of01d.o
hsl_ma77d.o
hsl_mc34d.o
hsl_mc78i.o
hsl_ma86d.o
hsl_ma87d.o
hsl_zb01i.o
hsl_mc68i.o
hsl_mc69d.o
hsl_mc64d.o
mc30d.o
mc77d.o
hsl_mc80d.o
hsl_ma97d.o
mc60d.o
mc61d.o
mc13d.o
mc20d.o
ma33d.o
hsl_zb01d.o
hsl_ma48d.o
hsl_mi28d.o
fa14d.o
fd15d.o
mc29d.o
la15d.o
la04d.o
and these object files contain the (dummy) references your linker claims are missing. I wonder if the order matters, i.e., perhaps libgalahad_hsl.a needs to occur before libgalahad.a?

nimgould:
You are right about the missing -lhwloc, I missed it as I compiled with dummy ssids rather than the real one.  But once I add this, the script forms the shared library under Linux.

amontoison:
> I wonder if the order matters, i.e., perhaps libgalahad_hsl.a needs to occur before libgalahad.a?

It could explain a lot of things because MacOS doesn't have `ranlib`.

dpo:
Yes I'm sure order matters, just as it does when compiling. However, I'm not finding the right order. The linux linker is too permissive.

dpo:
Actually, I think the problem runs deeper. Those `...MOD...` symbols do not appear in any of the libraries. They refer to the F90 modules. I tried adding `-I$GALAHAD/modules/$ARCH/double` to the final `gfortran` command, but that doesn't solve the problem.

It's important to realize that this way of generating shared libraries is just a patch. It's not the "right" way to generate them. I used it successfully on C and F77 projects, but I had never tried it on F90 libraries that depend on F90 modules. Perhaps it's simply insufficient.

amontoison:
> Actually, I think the problem runs deeper. Those `...MOD...` symbols do not appear in any of the libraries. They refer to the F90 modules. I tried adding `-I$GALAHAD/modules/$ARCH/double` to the final `gfortran` command, but that doesn't solve the problem.
> 
> It's important to realize that this way of generating shared libraries is just a patch. It's not the "right" way to generate them. I used it successfully on C and F77 projects, but I had never tried it on F90 libraries that depend on F90 modules. Perhaps it's simply insufficient.

We are doing the same thing for HSL packages Dominique, which are F90 libraries that sometimes depend on F90 modules, and it works.

dpo:
Not on macOS.

dpo:
@nimgould On a related topic, currently, it's hard to debug `create_one_shared` because it's called with `$(BINSHELL)` in `src/makemaster`. So adding `-vx` to the first line of the script has no effect (at least on macOS).

Since it's executable, it's enough to remove `$(BINSHELL)` like so:
```diff
diff --git a/src/makemaster b/src/makemaster
index fa5a7ed..f6d3d1b 100644
--- a/src/makemaster
+++ b/src/makemaster
@@ -1630,7 +1630,7 @@ create_one_shared:
        cd $(OBJ); CC=$(CC) FORTRAN=$(FORTRAN) OPTIMIZATION=$(OPTIMIZATION) \
                    SHARED=$(SHARED) DLEXT=$(DLEXT) LOADALL=$(LOADALL) \
                    LOADNONE=$(LOADNONE) \
-                   $(BINSHELL) $(GALAHAD)/bin/create_one_shared
+                   $(GALAHAD)/bin/create_one_shared

 #  book keeping
```
Then, `-vx` works as expected.

amontoison:
> More broadly, I now think we should provide two variables METIS4 and METIS5 in every compiler.* file in ARCHDEFS, setting as defaults
> 
> METIS4='-lgalahad_metis4' METIS5=
> 
> This will then permeate into the $GALAHAD/makefiles/* on install, and will provide users opportunities subsequently to replace these with real versions of metis4 and metis5 if they are available. Incidentally, metis5 (and parametis 4) is available as part of the Linux distributions, as is mumps with or without (p)scotch orderings. Is that true on Macs as well? I have always downloaded and compiled metis 4 as described in the README.external file in galahad. (same address as provided by @dpo above).

On Mac, MUMPS, SCOTCH, METIS, PARMETIS is available with Homebrew. It's Dominique that added them.
We would like to use [`Ygg`](https://github.com/fredrikekre/ygg) to easily provide the shared libraries precompiled by BinaryBuilder / Yggdrasil on all platforms with the same API.

> What do you think @dpo @jfowkes @amontoison ? If you agree, someone will have to make a global change to the archdefs files, and we need to coordinate the change with those for galahad so that they stay in sync

Good idea Nick, I created a PR for METIS4 and METIS5 variables in ARCHDefs repository.

nimgould:
OK, lots of things here.

1. Thanks @dpo for the change to the main makefile to use the correct shell.

2. Thank @amontoison for adding those variables. I will check and approve after I've grabbed a cup of tea. On this, since you mention it, could we also add
SCOTCH=
PARAMETIS=
MPI='-lgalahad_mpi"
the first two for future proofing (and because mumps uses the first), and the last
because mumps uses mpi, and if mumps isn't available I need to pass in  dummy mpi_init and mpi_terminate routines when building the interface to sls.

3. the Mod file issue. This is strange. A mod file is only required in fortran when one package uses(calls)  another, it is quite like a .h file in c (although of course a module is more than just that). The fact that the "complete" shared library is complaining about the hsl modules, but not the galahad ones suggests that this may not be the issue; I would expect complaints about all modules otherwise. Also, you say that you have julia interfaces to hsl, and hsl itself is module based, so I would have expected to see the same issue there.

 What might be possible is to build the shared libraries as the compilation proceeds. Can one add objects to a shared library, or does it all have to happen at once? (My ignorance, this is trivial for random libraries, indeed the whole point of them, and it is a shame that macos doesn't have such a useful tool ... or does it?)

dpo:
> What might be possible is to build the shared libraries as the compilation proceeds.

No, unfortunately, as far as I know, that is not possible (not on linux either).

nimgould:
OK, we need to find out why the process is not complaining about missing galahad mods but does for the hsl ones. Why does it not complain about hwloc when that is only available via its library? What about the other dummies like pastix? Is it simply that they are fortran 77 or c/c++. I really don't understand how macos shared libraries work, do you? You hinted that the hsl build doesn't work on macs on its own. Any idea why? Presumably it would be trivial to build a pair of 5 line modern fortran modules, one of which uses the other, to see this failure in it simplest form, and then to ask the question to the mac community?

nimgould:
Another possibility is to unpack all of the object files from the static libraries into a directory using ar x libname.a and then build the shared library from the .o files. If you want to try, there is another script , build_one_shared, and this can be tried using

 make -s -f $GALAHAD/makefiles/(yourarch) build_one_shared

and this produces a libgahad_all.so/dylib in 

 $GALAHAD/objects/(yourarch)/double/shared

Of course this does depend on ar working properly, but from what I see on google about Macs this is so. As I said before, I don't believe that the issue is not finding mod files, as there is nothing to suggest that it hasn't found the galahad ones. On linux,   

 nm -D -g libgalahad_all.so

tells you what is in the shared library, and everything seems to be there. I appreciate that this may be superseded by @dpo 's meson build

amontoison:
Do we want to keep this PR to test the `makemaster` build system?

PR #25, opened 2022-11-20
Use C_INT64_T instead of C_LONG_LONG

amontoison:
We also need to update the headers files, such that I can regenerate the Julia interface.


amontoison:
@nimgould @jfowkes 
I updated header files.
I probably need to revert the modifications of HSL headers, what do you think?

I also changed the type of some attributes with this [commit](https://github.com/ralna/GALAHAD/pull/25/commits/ffbcf1a29c8183026d9068e0ce7e660edd00b668), they were different from what I found in f90 files.

Is it normal that for some packages, `factorization_integer` and `factorization_real` are 64 bits int (`bqpb`, `ccqp`, `clls`, `cqp`,`dqp`,`fdc`,`sbls`) and for other packages they are 32 bits int (`arc`, `eqp`, `lpb`, `lsqp`, `nls`, `psls`, `qpa`, `qpb`, `trb`, `tru`, `wcp`)?

nimgould (approved):
Thanks, that is extremely useful!

nimgould:
Yes, this is a bit of a mess. factorization_integer/real in the packages that you highlight do nothing with these parameters, although they should (clearly, this was the intention, but overlooked). I'll have to work through these, and I agree that they should all be 64bit integers. Thanks

nimgould:
I have made all of the changes to provide factorization space data for the packages you mentioned. I'll push in an hour or so, I need to push something else from my machine at home first.

nimgould:
All done, and all seems fine in Linux-land

amontoison:
Perfect Nick, I will just open a small pull request to add `#include <stdint.h>` in all header files.
For the header files of HSL packages, do we have the latest versions in GALAHAD?

nimgould:
Not quite, there have been minor recent hsl updates, in particular incorrectly named c interfaces in a few cases. I'll sort this out today, and let you know when I'm done

PR #26, opened 2022-11-22
Meson build

dpo:
⚠️ **WORK IN PROGRESS** ⚠️ 

@amontoison @jfowkes @nimgould 

This is work in progress to demonstrate how far I got after one day of work trying to modernize the GALAHAD build system. I chose to use the `meson` build system as it's shaping up to be the post-CMake future. Its goals are to be fast and to make the build scripts easy to write, to read, and to understand.

I got a number of GALAHAD modules to build (those that don't require HSL, not even the dummy HSL). GLTR and GLRT are in there.

The bottom line is:
* the build is very fast;
* it's easy to set which compilers we want to use (`meson` knows about a number of C and Fortran compilers);
* the continuous integration script in this pull request demonstrates gcc/gfortran and icc/ifort on Linux and macOS;
* building dynamic libraries on both platforms was painless; (this is the number 1 benefit!)
* we no longer need ARCHDefs;
* we can have several "build trees" of GALAHAD simultaneously, each corresponding to compiler combinations (just as before);
* we can also install the executables and libraries (e.g., to `/usr/local/`), as if with `make install`.

There are other hidden benefits that future-proof GALAHAD, including creating source distributions, creating deb or rpm packages, etc.

Have a look at the main build script `meson.build` in the top directory. It could hardly be simpler. Happy to give you a tour during our next Zoom.

Still very much learning but I'm confident that I'll be able to add SSIDS, HSL, etc., to the builds.

It would be great if you could try it locally. There are instructions in `README.meson` (it's easy).


dpo:
### Status

- [ ] ~~all_go~~
- [ ] ~~all_qp~~
- [ ] ~~all_rq~~
- [ ] amd
- [ ] ampl
- [ ] apps
- [ ] arc
- [ ] basic
- [ ] bgo
- [ ] blls
- [ ] bqp
- [ ] bqpb
- [ ] bsc
- [ ] buildspec
- [ ] ccqp
- [ ] cdqp
- [ ] changes
- [ ] check
- [ ] checkpoint
- [ ] clls
- [ ] clock
- [ ] common
- [ ] convert
- [ ] copyright
- [ ] cqp
- [ ] cro
- [ ] cutest_functions
- [ ] demo
- [ ] dgo
- [ ] dlp
- [ ] dps
- [ ] dqp
- [ ] dum
- [ ] eqp
- [ ] extend
- [ ] external
- [ ] fdc
- [ ] fdh
- [ ] filter
- [ ] filtrane
- [ ] fit
- [ ] ~~forthcoming~~
- [ ] general
- [ ] glrt
- [ ] gls
- [ ] gltr
- [ ] hash
- [ ] hsl
- [ ] ir
- [ ] l1qp
- [ ] l2rt
- [ ] lancelot
- [ ] lancelot_simple
- [ ] lapack
- [ ] lhs
- [ ] lls
- [ ] llst
- [ ] lms
- [ ] lmt
- [ ] lpa
- [ ] lpb
- [ ] lpqp
- [ ] lqr
- [ ] lqt
- [ ] lsp
- [ ] lsqp
- [ ] lsrb
- [ ] lsrt
- [ ] lstr
- [ ] matlab
- [ ] miqr
- [ ] mop
- [ ] nlpt
- [ ] nls
- [ ] norms
- [ ] ~~oblivion~~
- [ ] opt
- [ ] presolve
- [ ] problem
- [ ] psls
- [ ] ptrans
- [ ] qp
- [ ] qpa
- [ ] qpb
- [ ] qpc
- [ ] qpd
- [ ] qpp
- [ ] qpt
- [ ] rand
- [ ] roots
- [ ] rpd
- [ ] rqs
- [ ] sbls
- [ ] scale
- [ ] scaling
- [ ] scu
- [ ] sec
- [ ] sha
- [ ] sils
- [ ] slls
- [ ] sls
- [ ] smt
- [ ] sort
- [ ] space
- [ ] specfile
- [ ] spral
- [ ] ssids
- [ ] string
- [ ] symbols
- [ ] tools
- [ ] trans
- [ ] trb
- [ ] trs
- [ ] tru
- [ ] ugo
- [ ] uls
- [ ] userdata
- [ ] warm
- [ ] wcp
- [ ] zd11

nimgould:
Although it looks like hsl is everywhere, it is really restricted to sils, sls, gls, uls and lpa (and mdch/precn in lancelot). So once you have those sorted out, the rest should be "straightforward"

dpo:
@nimgould The code lives in the `meson-build` branch, so all you should have to do is fetch all the branches from upstream. With git, that would be:
```
git fetch
git checkout meson-build
```

I haven't yet indicated dependencies between modules; only what goes into what library. Also I'm sure there will be a way to build the Python and Matlab interfaces. In a first stage, I wanted to focus on building standalone libraries that we can use in Julia. Oh, and we should also be able to build on Windows (I just need to figure out how to set up compilers on the CI virtual machines).

You'll be disappointed that `meson` doesn't really generate something readable. They parted with Makefiles because `make` was in large part responsible for slow builds. `meson` generates so-called "ninja" files. They vaguely look like Makefiles (they're text files). There will be one called `build.ninja` in the build directory if you try this locally. Ninja is responsible for the fast builds. However, you can build with `meson compile -C builddir -v` to see all the compile commands as they are executed.

A couple of questions:

* Is the SSIDS code in GALAHAD a copy of the original? Is it an option for the user to build SSIDS separately and tell GALAHAD where it is?
* Isn't SSIDS part of SPRAL? Why are there two GALAHAD modules?

nimgould:
@dpo Thank you, I will try tomorrow. As to your questions:

1. It is a copy, but only of the relevant parts. Getting the crummy cmake to coincide with the rest of the build, plus telling users to install something else, went against the intention to provide a standalone version with sparse matrix capabilities. If a user really wants to build their own, they simply replace the -lgalahad_spral with their library (although I admit I have never done this).
2. Yes it is, but spral comes with a lot of unnecessary baggage that I stripped out, and I didn't want anybody to think that this was the same as the official one. In retrospect our group felt that  it was a mistake to put ssids into spral, as the spral build brings in every other package, some of which have their own issues. That's the history, and it works fine as it stands, or at least as well as ssids does without real support. As you said before, ultimately we will replace this as the default with mumps ... but that has its own issues, particularly lacking all the functionality that I persuaded my colleagues to put into HSL/ssids (modified factors, inertia cntrol, Fredholm alternatives, triangular solves with sparse right-hand sides).

amontoison:
I tested your Meson build Dominique and it works fine on my computer. In less than 1 minute I was able to compile and test every packages supported by your Meson build script.

nimgould:
Wow, I didn't even get a chance to open the first beer, that was mighty fast! Was it optimizing the compilations? Whatever, it looks like a great tool, @dpo

dpo:
Great! I'll continue adding to it then. Currently, only double precision is supported, but yes, everything is compiled and tested in parallel. If you get a chance, try building with the NAG compiler (I don't have access to it). In `liblapack`, there's a conditional based on whether we're using that compiler or not (see `src/lapack/meson.build`). I would like to know if it's working.

I'm working on adding MA27 (and SILS). I have a few questions:

1. I wasn’t able to find what license MINPACK-2 is distributed under (MINPACK-1 on Netlib uses a permissive license). Do we really need it?
2. I can’t seem to download ICFS anymore; how important is it in GALAHAD?
3. I downloaded hslarchive-galahad. The ma33 directory contains mc13, 21, 33, and 20. Say a user downloads MA33 separately from the HSL Archive (perhaps b/c they already have MA27 and AD02). Is the user getting those same files? Or does hslarchive-galahad somehow package things differently? Same question about hslgalahad.
4. Is `sed` really the best/right way to switch from double to single precision in Fortran projects? In C/C++, I believe you would use the preprocessor and pass a compile flag such as `-Dsingle` or so. There is also a Fortran preprocessor.

nimgould:
I'm not sure how to say which compiler to use; presumably there is some build flag? Where are these listed, is there a file to see all options? Also, how do I turn on optimization, etc?

1. I will look, but really this isn't vital as I believe that gqt from MINPACK 2 stuff is only used in esoteric testing. I will check, and we can probably avoid this.
2. https://github.com/optimizers/lldl/tree/master/src/icf is one source ... yours. It is still a powerful weapon in the preconditioning armoury.  
3. Yes, the HSL package dump system is very primitive, and this is how every user will get it. Each package directory contains all of its dependencies, and thus there are multiple repetitions. It might be possible to change this, but as my colleagues are more interested with a different new-hsl future, I doubt there will be any enthusiasm to change something that sort of works. I once wrote a unified make system for hsl, but Jonathan hated it, and it never survived the cull. New HSL will of course embrace the 21st century!! 
4. That was mother and invention. Although there was an attempt to standardise a fortran preprocessor, it never took off as each vendor already had their own (different) one. Thus I simply chose sed as it avoided the vendor wars. Not ideal, of course, and one of the areas that C definitely wins. The most important thing, of course, is to avoid copies of files that only differ in a few places since that doubles maintenance, something that has been painful for hsl. 

ps, yes the generated ninja files are pretty verbose, but somehow not as bad as Cmake's dreadful makefiles

dpo:
Thanks @nimgould.

The ICFS README sounds quite permissive. Should we just bundle it inside GALAHAD?

I got GLS and SILS going. We'll have to decide if we only want to support hslarchive-galahad or if we also want to allow the user to specify AD02, MA27 and MA33 separately. That would add a lot of command-line options but is doable. Even more so with hslgalahad.

I think I should be able to do away with the dummy packages, and simply not build SILS (say) if MA27 isn't present. Does that sounds like a viable option?

Now that I've figured out how to add HSL dependencies, I should focus on SSIDS. I can already tell that that one will be fun.

nimgould:
I have sent Jorge More a message asking for permission to include ICFS. Just as with HSL, we need to be very careful what we say. Of course, if he says yes, I'd slightly rework it so that it is in modern fortran, and provide a C interface. But he might equally say no, and then we can make a decision on what to do (I see he has given explicit permission for another physics-based GPL-licenced project, 
and I have no idea what you negotiated for lldl). Minpack 1 seems to be BSD.

I believe that we should work on the assumption that people have or have not downloaded the (separate) HSL-acrchive and HSL bundles. There may be the occasional person who hasn't but really let's not worry.

I disagree about the dummy routines. We need these as otherwise the core linear solvers (sls, uls and psls) will not work, and then most of the rest of GALAHAD is stuck. Currently, we simply return a flag to tell an unaware user when a requested solver is not available. If, as I hope, ultimately GALAHAD appears both on individuals workstations/laptops but also on institute's supercomputers for which users will not necessarily be aware of which optional software is available.
It is for this reason that we need to keep HSL as a separate library, just as 
we do for LAPACK and the BLAS (and all the other external linear solvers).

I'm still mystified on how shared libraries work together. If elements of library libA
use library libB, then so long as I build libB first, can I then simply say -lB when building libA. This seems to be the case when we use -lstdc++ -lgomp -lhwloc; I am presuming that these libraries have all the symbols that are needed?

Thanks for the HOWTO re compilers. What about other flags? Can I add optimization (-O#), can I add debug (-g etc)?

dpo:
Thanks for enquiring about ICFS.

I see what you're saying about the linear solvers. I'll keep the dummy routines in there.

Certain dynamic libraries can be "incomplete" in the sense that the linker will figure out whether it's finding all symbols or not (the `--allow-undefined` flag lets you build such an incomplete library; on macOS, the equivalent flag is `-undefined dynamic_lookup`). You can't use an incomplete library on its own obviously; you must also link whatever other libraries define the missing symbols. To answer your question, yes, you can say `-lB` when building `libA`. Here, `libA` would not be "incomplete" because it links explicitly to `libB`. However, I believe you could also build `libA` with `--allow-undefined`, but then users would have to also link explicitly against `libB` when they try to use `libA`.

nimgould:
I have removed all references to minpack's gqt; it was an unnecessary add-on. So it is only icfs that we need to sort out.

Thanks for your views on shared libraries, that all makes sense.

dpo:
@nimgould I added two features to the meson build script:

1. users can specify BLAS and LAPACK libraries (if none is specified, the built-in implementation is used);
2. dummy AD02, MA27 and MA33 are built into libgalahad_hsl.so if the user doesn't provide the path to hslarchive-galahad.

When I ran the example programs or test executables for GLS and SILS with the dummy routines, I expected to see the message saying "we're sorry that ...", but instead only saw a message saying that the symbolic analysis failed. Is that expected?

nimgould:
Yes, that is to be expected. The failure message is disabled by default so that the "no printing" 
rule stipulated by many applications is observed. I have changed the output message from the test program as the status value overflowed the ancient format.

Given the compile times, I doubt that debug is enabled by default as this tends to be rather slow on Linux with gfortran.

I tried the latest version, but get a 

meson.build:1:0: ERROR: Meson version is 0.61.2 but project requires >= 0.64.0

I then tried a new install from the meson site using pip, but there I get

% pip3 install --user meson
Requirement already satisfied: meson in /usr/lib/python3/dist-packages (0.61.2)

I'm sure I can install from source if necessary, but I doubt users will so we need to be careful not to be too bleeding edge. Would you expect this?

amontoison:
I  also have the version `0.61.2` !

nimgould:
git clone https://github.com/mesonbuild/meson ./meson
gets the latest (0.64) version, and ./meson/meson.py runs everything.
Seems fine; apparently compiler uses default flags 
-D_FILE_OFFSET_BITS=64 -c -O0
and these show a number of unused variables that I have fixed

For nagfor I get

$ FC=nagfor /usr/local/src/meson/meson.py setup nag -Dciface=true
The Meson build system
Version: 0.99.99
Source dir: /home/nimg/fortran/meson/galahad
Build dir: /home/nimg/fortran/meson/galahad/nag
Build type: native build
Project name: GALAHAD
Project version: 5.0.0

meson.build:1:0: ERROR: Value "legacy" (of type "string") for combo option "Fortran language standard to use" is not one of the choices. Possible choices are (as string): "none". 

I can get it to create the "nag" build directory if I comment out    'fortran_std=legacy',
but then 
%~/fortran/meson/galahad $ /usr/local/src/meson/meson.py compile -C nag

INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /bin/ninja -C /home/nimg/fortran/meson/galahad/nag
ninja: Entering directory `/home/nimg/fortran/meson/galahad/nag'
[8/365] Compiling Fortran object libgalahad_blas.so.p/src_lapack_blas.f.o
FAILED: libgalahad_blas.so.p/src_lapack_blas.f.o 
nagfor -Ilibgalahad_blas.so.p -I. -I.. -quiet -O0 -g -PIC -mdir libgalahad_blas.so.p -o libgalahad_blas.so.p/src_lapack_blas.f.o -c ../src/lapack/blas.f
Non-standard(Obsolete): ../src/lapack/blas.f, line 5590: DOUBLE COMPLEX keyword
                        detected at <end-of-statement>@DOUBLE COMPLEX
Non-standard(Obsolete): ../src/lapack/blas.f, line 5594: DOUBLE COMPLEX keyword
                        detected at <end-of-statement>@DOUBLE COMPLEX
Non-standard(Obsolete): ../src/lapack/blas.f, line 5633: DOUBLE COMPLEX keyword
                        detected at <end-of-statement>@DOUBLE COMPLEX
Non-standard(Obsolete): ../src/lapack/blas.f, line 5883: DOUBLE COMPLEX keyword
                        detected at <end-of-statement>@DOUBLE COMPLEX
Non-standard(Obsolete): ../src/lapack/blas.f, line 5888: DOUBLE COMPLEX keyword
                        detected at <end-of-statement>@DOUBLE COMPLEX
Non-standard(Obsolete): ../src/lapack/blas.f, line 6022: DOUBLE COMPLEX keyword
                        detected at <end-of-statement>@DOUBLE COMPLEX
Non-standard(Obsolete): ../src/lapack/blas.f, line 6027: DOUBLE COMPLEX keyword
                        detected at <end-of-statement>@DOUBLE COMPLEX
Non-standard(Obsolete): ../src/lapack/blas.f, line 6029: DOUBLE COMPLEX keyword
                        detected at <end-of-statement>@DOUBLE COMPLEX
Error: ../src/lapack/blas.f, line 6294: DCONJG is not the name of an intrinsic procedure
Error: ../src/lapack/blas.f, line 12006: DIMAG is not the name of an intrinsic procedure
Non-standard(Obsolete): ../src/lapack/blas.f, line 12020: Byte count on numeric data type
                        detected at *@16
and a lot of similar. nagfor uses -fixed to specify fixed-format f77 codes (like lapack/blas), and there is an additional -dusty flag (that I haven't used in the current galahad make). But crucially, since DCONJG and DIMAG are not standard fortran, nagfor requires the -dcfuns flag. 

I tied putting this into build.ninja, and got further, now only warnings from blas and lapack compiles, but then 
nagfor -Ichecks.p -I. -I.. -Ilibgalahad.so.p -quiet -O0 -g -mdir checks.p -o checks.p/src_check_checks.f90.o -c -dcfuns ../src/check/checks.f90
Fatal Error: ../src/check/checks.f90, line 5: Cannot find module GALAHAD_BLAS_INTERFACE
             detected at GALAHAD_NLPT_DOUBLE@<end-of-statement>
and a whole bunch of other similar ones.

I wonder ... in the old system, the blas_interface and lapack_interface modules are part of the galahad shared library, not the blas/lapack ones (they simply provide generic interfaces to the actual blas/lapack routines). When I look in .ninja.log, I see
1	23	1669372657816830833	libgalahad_blas.so.p/galahad_blas_interface.mod	c1a2424c5b6380aa
which might suggests that the new build is putting them into the wrong binary library. Just speculation. However. the gfortran install didn't choke. I also noticed that in the wild flurry of parallel compilations, there didn't seem to be a compile of these _interface packages, so maybe you simply missed them; there is no mention of them in mason.build

amontoison:
We need to set `-Ddebug` and `-Doptimization`  for the other flags Nick. I just checked the documentation.
```
...
  Core options        Current Value        Possible Values                                               Description
  ------------        -------------        ---------------                                               -----------
...
  debug               false                [true, false]                                                 Debug
...
  optimization        3                    [0, g, 1, 2, 3, s]                                            Optimization level
...
```

dpo:
Try commenting out the version restriction at the top of `meson.build`. I'll look into the rest later. Yes, I'm putting the blas/lapack `_interface.f90` files in `libgalahad_blas`.

dpo:
@nimgould I rebased your latest changes, but I'm now getting:
```
[169/365] Compiling Fortran object gltrs.p/src_gltr_gltrs.f90.o
FAILED: gltrs.p/src_gltr_gltrs.f90.o
gfortran -Igltrs.p -I. -I.. -Ilibgalahad.dylib.p -fdiagnostics-color=always -Wall -O0 -g -Jgltrs.p -o gltrs.p/src_gltr_gltrs.f90.o -c ../src/gltr/gltrs.f90
../src/gltr/gltrs.f90:25:11:

   25 |        DO i = 2, n - 1
      |           1
Error: Symbol 'i' at (1) has no IMPLICIT type
```
Looks like that `i` was being used.

dpo:
To complete @amontoison's screenshot, we see in the manual that the default build is indeed a debug build:

Screen Shot 2022-11-25 at 9 27 28 PM

cf https://mesonbuild.com/Builtin-options.html#core-options

amontoison:
> @nimgould I rebased your latest changes, but I'm now getting:
> 
> ```
> [169/365] Compiling Fortran object gltrs.p/src_gltr_gltrs.f90.o
> FAILED: gltrs.p/src_gltr_gltrs.f90.o
> gfortran -Igltrs.p -I. -I.. -Ilibgalahad.dylib.p -fdiagnostics-color=always -Wall -O0 -g -Jgltrs.p -o gltrs.p/src_gltr_gltrs.f90.o -c ../src/gltr/gltrs.f90
> ../src/gltr/gltrs.f90:25:11:
> 
>    25 |        DO i = 2, n - 1
>       |           1
> Error: Symbol 'i' at (1) has no IMPLICIT type
> ```
> 
> Looks like that `i` was being used.
> 
> PS: still getting quite a number of unused variables/arguments: https://gist.github.com/871d09c114faf8d0e9daa264f8244019.
> 
> PPS: For me, the default compilation options seems to be: `-O0 -g`, so: a debug build.

I have different unused variables here: https://gist.github.com/amontoison/c2781e0946418a617fbdeeca660bf125
I use Meson 0.61.2 on my computer for information.

dpo:
@nimgould @amontoison I rebased a made the following changes:
- the min version is now 0.61 (arbitrarily, since that's the one you get on Linux);
- removed the `fortran_legacy` bit;
- put f90 BLAS/LAPACK interfaces in `libgalahad` instead of `libgalahad_blas/lapack`;
- added extra flags for nagfor if building our own BLAS/LAPACK.

Apparently, it's possible to download a trial version of nagfor. I'll do that so I can test locally. Locally, gfortran and ifort are happy.

meson.build
```diff
@@ -316,7 +318,7 @@ test('lms', lmst)
 miqrt = executable('miqrt', 'src/miqr/miqrt.f90', link_with : libgalahad)
 test('miqr', miqrt)
 
-mopt = executable('mopt', 'src/mop/mopt.f90', link_with : [libgalahad_blas, libgalahad], dependencies : blas_dep)
+mopt = executable('mopt', 'src/mop/mopt.f90', link_with : libgalahad_link_with + libgalahad, dependencies : blas_dep)
```
amontoison, line 321: I'm not sure to understand this line. If we want to use `libopenblas` instead of `libgalahad_blas` and `libgalahad_lapack`, we still need to link `mopt` with it?

dpo: In that case, `libgalahad_blas/lapack` don't appear in `libgalahad_link_with`, but `openblas` appears in `blas_dep`. With Meson, `link_with` is reserved for targets that we build ourselves, while `dependencies` are for external libraries. If we build our own BLAS, then `blas_dep` is empty.

amontoison: Perfect, thanks for the clarification!

nimgould:
Thanks for comments regarding unused variables. The obvious ones have now been removed (and the one @dpo mentioned re-instated). Some are beyond our control (e.g, lapack and hsl ones), some others are placeholder subroutine arguments that may be used one day.

nimgould:
Odd that it claims to give optimization and debug builds as the command flags I see are -O1 and a very basic debug without any runtime checks

nimgould:
nag build now tries to use -fixed-dcfuns without a space between the two, and then fails.

A git issue ... I made a local change, then tried to pull and git complained. I then tried to tell it to ignore my changes (using its suggestion of "git config pull.rebase false"), but it then get into a horrible mess, telling me that I should stash file before I pull, but when it did it said that I couldn't because I had unresolved conflicts ...  and as Ididn't know what was wrong, I had to reinstall. With svn reconciling such things is easy, what should I do with git? How do I say just replace anything local with global changes?

dpo:
Thanks @nimgould. I tried to fix the NAG build flags. Does it work now?

The git issue is probably because I'm "force-pushing" updates to this branch each time I rebase with the master branch (which means that I reapply this branch's changes on top of the updates you are making to the master branch). So this branch looks different each time. The easiest is probably to delete your local meson-build branch, `git fetch`, and check out the branch as if it were the first time.

eli-schwartz:
> ps, yes the generated ninja files are pretty verbose, but somehow not as bad as Cmake's dreadful makefiles

Part of that is because ninja itself has some nice advantages over Make that require some ridiculous contortions to get Make to do at all. Things like build edges depending on the compiler rule that produced it, so that reconfiguring the build only forces you to recompile source files whose compilation options actually changed.

meson.build
```diff
@@ -0,0 +1,380 @@
...
+  if fs.is_dir(hslarchive_galahad)
+    ad02_path = hslarchive_galahad / 'ad02'
+    ma27_path = hslarchive_galahad / 'ma27'
+    ma33_path = hslarchive_galahad / 'ma33'
+    has_hslarchive = true
+  else
+    error('directory not found: ', hslarchive_galahad)
+  endif
+else
+  warning('consider downloading hslarchive-galahad to improve functionality')
```
eli-schwartz, line 84: You may want to consider using https://mesonbuild.com/Wrap-dependency-system-manual.html#specific-to-wrapfile to make this easier.

It does look like hslarchive-galahad requires an agreement to download, so you cannot take advantage of automatic wrap dependency downloads, but you can give people a location to download it to (`subprojects/packagefiles/`) and also abstract out the logic for building it into an overlay meson.build

src/amd/meson.build
```diff
@@ -0,0 +1,2 @@
+amd_src = ['src/amd/amd.f90']
+libgalahad_src += amd_src
```
eli-schwartz, line 2: In all these subdirectories that add files to the main sources list, you could do something like this instead:

```meson
libgalahad_src += files('amd.f90')
```

Files objects remember where they were defined, so you don't have to sync the preceding path relative to the root of the git repo everywhere. This also means that if, in the future, you move some directories around, you don't need to edit these strings.

dpo:
@nimgould I tried the NAG compiler locally, and all seems well after the most recent updates to this branch.

src/amd/meson.build
dpo: Thanks! Very much still learning meson.

nimgould:
Yes, the nagfor version compiles perfectly ... and flags a whole new bunch of unused variables to remove at some stage. Thanks

amontoison:
@eli-schwartz A quick question for you, do we recommend to keep all the tests and examples in the main `meson.build` or to move them in the `meson.build` of each package?

eli-schwartz:
```meson
tests = [
    ['single', extra_args_single, libgalahad_single],
    ['double', extra_args_double, libgalahad_double]
]

foreach t: tests
    exe = executable ('amdt_'+t[0], 'src/amd/amdt.F90', fortran_args : t[1], link_with : t[2], include_directories: libgalahad_include)
    test('amd_'+t[0], exe, suite: ['fortran', t[0]], is_parallel : false)
endforeach
```

On a personal level I would tend to use subdir() and then define tests/examples in their own meson.build file, yes.

amontoison:
Thanks @eli-schwartz!

amontoison:
@eli-schwartz 
I have two new questions for you if you find time to help us:
1- Is it normal that in some cases, files are not compiled in the same order that we added them in a `files` structure? 
```
libmumps_src += files('src/kinds/kinds.F90', 'src/dum/mpi.F90', 'src/dum/mumps.F90')

libmumps_single = library('mumps_single',
                          sources : libmumps_src,
                          link_with : [libgalahad_blas, libgalahad_lapack],
                          dependencies : blas_dep + lapack_dep,
                          fortran_args : extra_args_single,
                          include_directories: libgalahad_include + 'src/dum/include',
                          install : true)

libmumps_double = library('mumps_double',
                          sources : libmumps_src,
                          link_with : [libgalahad_blas, libgalahad_lapack],
                          dependencies : blas_dep + lapack_dep,
                          fortran_args : extra_args_double,
                          include_directories: libgalahad_include + 'src/dum/include',
                          install : true)
```
```
[120/2062] Compiling Fortran object libmumps_single.so.p/src_kinds_kinds.F90.o
[121/2062] Generating symbol file libmkl_pardiso.so.p/libmkl_pardiso.so.symbols
[122/2062] Generating symbol file libpardiso.so.p/libpardiso.so.symbols
[123/2062] Compiling Fortran object libmumps_single.so.p/src_dum_mpi.F90.o
[124/2062] Compiling Fortran object libmumps_single.so.p/src_external_mumps_mumps_types.F90.o
[125/2062] Compiling Fortran object libmumps_double.so.p/src_dum_mpi.F90.o
FAILED: libmumps_double.so.p/src_dum_mpi.F90.o 
gfortran -Ilibmumps_double.so.p -I. -I.. -I../include -I../src/dum/include -Ilibgalahad_lapack.so.p -Ilibgalahad_blas.so.p -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -O0 -g -fPIC -DGALAHAD_DOUBLE -DSPRAL_DOUBLE -DSPRAL_NO_HWLOC -DSPRAL_NO_SCHED_GETCPU -Jlibmumps_double.so.p -o libmumps_double.so.p/src_dum_mpi.F90.o -c ../src/dum/mpi.F90
../src/dum/mpi.F90:8:8:

    8 |    USE GALAHAD_KINDS_precision
      |        1
Fatal Error: Cannot open module file ‘galahad_kinds_double.mod’ for reading at (1): No such file or directory
```
Because `src/kinds/kinds.F90` is the first file in the `files` object, I expected that it's the first one compiled.
Do you know how to solve to problem?

2-  We would like to merge the single and double precision libraries, which means that we will compile two times the same files but with different flags for one library. Do you know how to do that?
With a basic makefile, I know that we need to add `_precision` at the end the object files to not overwrite object files compiled for a different precision but I don't know if Meson handles it automatically.

eli-schwartz:
> 1- Is it normal that in some cases, files are not compiled in the same order that we added them in a `files` structure?

In all cases, really. The files structure is just an arbitrarily unsorted list, and ninja tries to compile as many things at the same time as it can. So it will compile all the sources for mumps_double at the same time if it can -- and that's actually fine, they shouldn't depend on each other.

What they do depend on is one single file:

```
gfortran -Ilibmumps_double.so.p -I. -I.. -I../include -I../src/dum/include -Ilibgalahad_lapack.so.p -Ilibgalahad_blas.so.p -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -O0 -g -fPIC -DGALAHAD_DOUBLE -DSPRAL_DOUBLE -DSPRAL_NO_HWLOC -DSPRAL_NO_SCHED_GETCPU -Jlibmumps_double.so.p -o libmumps_double.so.p/src_dum_mpi.F90.o -c ../src/dum/mpi.F90
../src/dum/mpi.F90:8:8:

    8 |    USE GALAHAD_KINDS_precision
      |        1
Fatal Error: Cannot open module file ‘galahad_kinds_double.mod’ for reading at (1): No such file or directory
```

mpi.F90 is importing a module, so it needs the .mod file here, right? We do have a dyndeps scanner for that, which runs once per target or so:

```
[1/2062] Module scanner.
[2/2062] Module scanner.
[3/2062] Module scanner.
[4/2062] Module scanner.
[5/2062] Module scanner.
[6/2062] Module scanner.
[7/2062] Module scanner.
[8/2062] Module scanner.
[9/2062] Module scanner.
[10/2062] Module scanner.
[11/2062] Module scanner.
[12/2062] Module scanner.
[13/2062] Compiling Fortran object libgalahad_blas.so.p/src_lapack_blas.f.o
[14/2062] Linking target libgalahad_blas.so
[15/2062] Module scanner.
[16/2062] Module scanner.

[...]
```

Those "Module scanner" rules are supposed to augment the interdependency links and thereby ensure that all fortran sources which e.g. import some module only get compiled after the .mod interface file exists. You can see our depscan implementation here: https://github.com/mesonbuild/meson/blob/master/mesonbuild/scripts/depscan.py

From a quick look at this project it *seems* like your USE import is importing a name that's affected by the preprocessor? I didn't know that was possible, although when I think about it I suppose since it's not the preprocessor doing the import (unlike with C/C++) I suppose that is legal...

... I'm not sure what a good way to parse this would be. Seems like the only way to figure out what module it needs is to actually try compiling it, at which point it is too late.

One possible hack might be to run the preprocessor on the `*.F90` file to handle this textual replacement and then allow the depscanner to scan the processed file?

I could be completely off-track though -- I'm not much of a fortran person.

> 2- We would like to merge the single and double precision libraries, which means that we will compile two times the same files but with different flags for one library. Do you know how to do that?

To compile with different flags, you should compile some files in a `static_library()` target with different flags and then merge them together using `objects: internal_staticlib.extract_all_objects()` to add the individual objects of one target to another target.

nimgould:
The module-based compilation of modern fortran libraries requires that for any module d that uses modules b and c, b and c are compiled first. Modules b and c might depend on a module a. Thus the compilation agenda is that of a directed acyclic graph (DAG). Some fortran compilers provide a tool that prescans all the modules to provide the graph (and thus generate a makefile). It seems that what is needed here is a parallel compilation agenda that respects the DAG in meson. Is this possible?

nimgould:
Some of these issues ... and solutions ... are addressed in
 https://aoterodelaroza.github.io/devnotes/modern-fortran-makefiles/

nimgould:
... and there are a whole bunch of other similar tools, see https://fortranwiki.org/fortran/show/Build+tools
This mentions meson, and specifically says "Meson should be able to handle Fortran module dependencies and compile the modules in the correct order."

jfowkes:
@nimgould @amontoison the issue here is not that meson cannot handle the module dependency tree and create a parallel compilation agenda, that is precisely what the dependency scanner that Eli mentions does and this works exactly as intended.

The issue here is that meson's dependency scanner checks the use statements in each Fortran file and we have:
```Fortran
USE GALAHAD_KINDS_precision
```
so the dependency scanner tries to add `galahad_kinds_precision.f90` to the compilation agenda rather than the preprocessed `galahad_kinds_double.f90` (or `galahad_kinds_single.f90`). That is, as Eli says:

> Your USE import is importing a name that's affected by the preprocessor? I didn't know that was possible, although when I think about it I suppose since it's not the preprocessor doing the import (unlike with C/C++) I suppose that is legal...
> 
> ... I'm not sure what a good way to parse this would be. Seems like the only way to figure out what module it needs is to actually try compiling it, at which point it is too late.
> 
> One possible hack might be to run the preprocessor on the *.F90 file to handle this textual replacement and then allow the depscanner to scan the processed file?

We really are pushing the boundaries of the Fortran standard here!

nimgould:
Eli suggested "One possible hack might be to run the preprocessor on the *.F90 file to handle this textual replacement and then allow the depscanner to scan the processed file?"

That may be the way out, or at least for the compilers we currently support. We have

gfortran: -cpp -E
ifort: -fpp -EP
nagfor: -fpp -F

option to output the preprocessed file without compilation. But really what one would like is that detection of USE GALAHAD_x_precision prompts compilation of both the x_single and x_double modules before trying anything else.

nimgould:
This link is interesting - 
 https://mathstuf.fedorapeople.org/fortran-modules/fortran-modules.html
it suggests that the cmake system does allow preprocessing before it decides on its compile order. I came across this from
 https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5562
in which an experimental ninja project tries to use the same ideas for C20++ codes. Wouldn't it be wonderful if meson had this capability

jfowkes:
@amontoison we have now fixed the `GALAHAD_KINDS_precision` issues in the `multi-precision` branch by adding a dummy kinds_precsion to the two `kinds.F90` files (galahad and ssids). While this might not be the cleanest it does work.

We also had to add `src/spral/kinds.F90` to `meson.build` for ssids as well.

amontoison:
> @amontoison we have now fixed the `GALAHAD_KINDS_precision` issues in the `multi-precision` branch by adding a dummy kinds_precision module to the two `kinds.F90` files (galahad and spral). While this might not be the cleanest it does work.
> 
> We also had to add `src/spral/kinds.F90` to `meson.build` for ssids as well.

Thanks @jfowkes 
But we will probably have the same problem for the other USE statements of all packages / dependencies. I am wrong?

jfowkes:
@amontoison nope it works, we even pass all the tests under linux:
https://github.com/ralna/GALAHAD/actions/runs/4135396883/jobs/7147789180
with a few other changes we had to make!

amontoison:
> @amontoison nope it works, we even pass all the tests under linux: https://github.com/ralna/GALAHAD/actions/runs/4135396883/jobs/7147789180 with a few other changes we had to make!

Ok fantastic!
@nimgould @nimgould @jfowkes 
I created a branch `galahad5`, such that I can merge a little bit `meson-build` and `multi-precision` branches.
Nick is working on Meson today but he is not using my last modifications.
Let's work on the `multi-precision` branch now together.

nimgould:
Really, you shouldn't let me anywhere near git. I'm a one branch man!

jfowkes:
Yes we tried to make them to meson.build but the CI wouldn't run for the changes 😢

amontoison:
Guys, everything is repaired :toolbox: 
All our modifications are now on the branch `galahad5`.
Nick and Jari, you need to delete the old `multi-precision` branch.
```
git fetch origin
git checkout galahad5
git branch -D multi-precision # remove the branch
git checkout -b multi-precision # create a new one from galahad5
git rebase origin/multi-precision
```

nimgould:
We need to add all of the packages (arc, tru, trb, nls, bgo, dgo) that depend on psls to the meson build now that psls is free of lancelot

jfowkes:
@nimgould looks like they've already been added to the newly update `multi-precision`  branch:
https://github.com/ralna/GALAHAD/blob/multi-precision/meson.build
There are now some new test failures under Linux:
https://github.com/ralna/GALAHAD/actions/runs/4136671250/jobs/7150853182

nimgould:
Good

The current meson.build has (as far as I can see) not included the 
, env : ['GALAHAD=/home/runner/work/GALAHAD/GALAHAD']
clauses we added to the rpb tests yesterday, hence those failures.

I have no idea about the other failures, they all work here.

nimgould:
OK, looking closer, make is ignoring error 134 returns from c++; these must come from ssids (as this is the only c++ code). I will look closer, but am not particularly optimistic; maybe I can change the tests to use an LAPACk routine instead? I wonder why these weren't flagged yesterday?

jfowkes:
Are you also seeing this locally? Yes testing with LAPACK would certainly be more robust...

nimgould:
I'm seeing it if I tell make to stop rather than continue if there are error flags.

However, I have now narrowed it to a fortran allocate statement failure,  and this
looks like a compiler bug.  I'm using gcc-12 at home, while yesterday it was gcc-11 at work

jfowkes:
Interesting, the CI is using gcc-10.4.0 for the latest failed runs whereas for the runs that worked yesteday the CI was using gcc-11.3.0 (@amontoison changed the way gcc is installed).

PR #29, opened 2022-11-27
Replace all clong by C_INT64_T in Fortran 90 files

amontoison:
@nimgould @jfowkes 
Do I also nee to update these lines:
`INTEGER, PARAMETER :: long = SELECTED_INT_KIND( 18 )` ?


nimgould:
Although it is unlikely, the fortran language will return a -ve value of 64 bit integers are not supported, and this will case INTEGER (KIND = C_INT64_T) to fail. The alternative via the long as above is guaranteed to work as it will drop to 32 bit integer if that is all that is available.

jfowkes:
@amontoison that's not required as those lines explicitly define a 64bit long, see:

> SELECTED_INT_KIND(R) return the kind value of the smallest integer type that can represent all values ranging from -10^R (exclusive) to 10^R (exclusive). If there is no integer kind that accommodates this range, SELECTED_INT_KIND returns -1.

@nimgould while theoretically possible all current platforms support 64bit integers.

nimgould:
Agreed, although, old fool that I am, I still have a 32bit laptop that I keep as it is tiny and easy to carry around

jfowkes:
Ah yes, I forgot about legacy devices, good point

nimgould:
To be honest, I am not sure we should worry too much about such devices. We old fools hold back progress, and deserve what we get!

amontoison:
What is the issue with 32 bits computers?
They also support `C_INT64_T`, no?

nimgould:
Possible, possibly not. It would certainly only be in software if at all. The C standard was pretty vague at that time about short, int and long meant, and certainly there wasn't anything 
other than a single integer type in fortran, which was usually 32 bits. My old laptop is entirely
32 bit. You youngsters take this all for granted!

PR #33, opened 2022-12-06
Remove blanks and tabs at the end of the lines

amontoison:


amontoison:
We will take into account this

PR #70, opened 2023-01-13
Update extension of lancelot_steerings

amontoison:
I also removed duplicate and broken examples `eqps.options` and `qpas.options`.


nimgould (commented):
OK, thanks

PR #85, opened 2023-02-09
GALAHAD 5.0

amontoison:


amontoison:
@dpo @jfowkes @nimgould 
`galahad5` is the (backup) branch with all our current modifications for GALAHAD 5.
`multi-precision` is the branch of Nick such he only needs the commands
```
git pull
git commit -m "blabla"
git push
```
If we want to add a modification, I propose to PR for this branch (`multi-precision`) such that Nick never needs to invoke a `git rebase`.

nimgould:
Excellent. Simplicity is best. I am now wearing my dunce's hat for the rest of the month as a penance

jfowkes:
@nimgould looking good, only CCQP and EQP failing the tests now

dpo:
Hi guys. I just tried to compile GALAHAD from this branch with clang/clang++/nagfor. On macOS, I have to use the command
```shell
CC=clang CXX=clang++ FC=nagfor FFLAGS=-fpp meson setup builddir_nag
```
because `nagfor` does not preprocess by default (so it returns an error at the first `#ifdef`).

Compilation failed in the chain of headers `#include`d in `src/ssids/wrappers.cxx`. One of them is `#include <version>`. Normally, the angle brackets mean that the compiler should include a system header, not a project header. However, `clang++` included the file `version` that is in the GALAHAD source tree, which, of course, caused an error. This looks like a compiler bug to me. However, do we really need the `version` file in `GALAHAD`? If so, we should rename it `GALAHAD_version` or something similar to disambiguate.

I removed `version` and started over. Compilation then fails in `src/pardiso/mkl_pardisot.F90`, where the compiler complains that variables from `GALAHAD_KINDS` are being redefined. (There is a `USE GALAHAD_KINDS_precision` in there.) However, commenting out that line results in errors of the form
```
Error: ../src/external/mkl_pardiso/mkl_pardisot.F90, line 44: RP_ is not a constant
       detected at _@RP_
```

The combination icc/icpc/ifort fails with
```
FAILED: wsmpt_single.p/src_external_wsmp_wsmpt.F90.o
ifort -Iwsmpt_single.p -I. -I.. -I../include -Ilibwsmp.dylib.p -warn general -warn truncated_source -g -traceback -O0 -g -DGALAHAD_SINGLE -DSPRAL_SINGLE -DSPRAL_NO_HWLOC -DSPRAL_NO_SCHED_GETCPU -module wsmpt_single.p -gen-dep=wsmpt_single.p/src_external_wsmp_wsmpt.F90.o -gen-depformat=make -o wsmpt_single.p/src_external_wsmp_wsmpt.F90.o -c ../src/external/wsmp/wsmpt.F90
../src/external/wsmp/wsmpt.F90(7): error #7013: This module file was not generated by any release of this compiler.   [GALAHAD_KINDS]
  USE GALAHAD_KINDS_single
------^
```
and similar errors in `mkl_pardisot`, and `pastixt`.

With gcc-12/g++-12/gfortran-12, the build goes through. The `rpdt` tests failed because I forgot to define the `GALAHAD` environment variable (there should be an error message if it's not defined). All tests passed after I defined it.

nimgould:
Many thanks, Dominique. Just a few observation before Friday night gets in the way.

I couldn't find the reference to #include <version>. I must be looking in the wrong place.
Where is this?

You definitely need the USE GALAHAD_KINDS_precision. Which variables are being redefined (that is possible, of course, and should be corrected)?

the ifortran/icc works fine under linux. I don't understand the USE GALAHAD_KINDS_single not found, as this module is part of src/kinds/kinds.F90

dpo:
Hi @nimgould. The `#include` is on line 8 of `ssids/wrappers.cxx`:
```c++
#include <stdexcept>
```
which results in a chain of `#include`s until, at least on macOS, an `#include <version>` in a system header.

Here are the symbols the NAG compiler complains about:
```
[1/1851] Compiling Fortran object pardisot_single.p/src_external_pardiso_pardisot.F90.o
FAILED: pardisot_single.p/src_external_pardiso_pardisot.F90.o
nagfor -Ipardisot_single.p -I. -I.. -I../include -Ilibpardiso.dylib.p -quiet -O0 -g -fpp -DGALAHAD_SINGLE -DSPRAL_SINGLE -DSPRAL_NO_HWLOC -DSPRAL_NO_SCHED_GETCPU -mdir pardisot_single.p -o pardisot_single.p/src_external_pardiso_pardisot.F90.o -c ../src/external/pardiso/pardisot.F90
Error: ../src/external/pardiso/pardisot.F90, line 5: Redeclaration of symbol REAL_BYTES_ from USEd module GALAHAD_KINDS
       detected at GALAHAD_KINDS_SINGLE@<end-of-statement>
Error: ../src/external/pardiso/pardisot.F90, line 5: Redeclaration of symbol RP_ from USEd module GALAHAD_KINDS
       detected at GALAHAD_KINDS_SINGLE@<end-of-statement>
Error: ../src/external/pardiso/pardisot.F90, line 5: Redeclaration of symbol CP_ from USEd module GALAHAD_KINDS
       detected at GALAHAD_KINDS_SINGLE@<end-of-statement>
Error: ../src/external/pardiso/pardisot.F90, line 5: Redeclaration of symbol RPC_ from USEd module GALAHAD_KINDS
       detected at GALAHAD_KINDS_SINGLE@<end-of-statement>
[2/1851] Compiling Fortran object mkl_pardisot_single.p/src_external_mkl_pardiso_mkl_pardisot.F90.o
FAILED: mkl_pardisot_single.p/src_external_mkl_pardiso_mkl_pardisot.F90.o
nagfor -Imkl_pardisot_single.p -I. -I.. -I../include -Ilibmkl_pardiso.dylib.p -quiet -O0 -g -fpp -DGALAHAD_SINGLE -DSPRAL_SINGLE -DSPRAL_NO_HWLOC -DSPRAL_NO_SCHED_GETCPU -mdir mkl_pardisot_single.p -o mkl_pardisot_single.p/src_external_mkl_pardiso_mkl_pardisot.F90.o -c ../src/external/mkl_pardiso/mkl_pardisot.F90
Error: ../src/external/mkl_pardiso/mkl_pardisot.F90, line 5: Redeclaration of symbol REAL_BYTES_ from USEd module GALAHAD_KINDS
       detected at GALAHAD_KINDS_SINGLE@<end-of-statement>
Error: ../src/external/mkl_pardiso/mkl_pardisot.F90, line 5: Redeclaration of symbol RP_ from USEd module GALAHAD_KINDS
       detected at GALAHAD_KINDS_SINGLE@<end-of-statement>
Error: ../src/external/mkl_pardiso/mkl_pardisot.F90, line 5: Redeclaration of symbol CP_ from USEd module GALAHAD_KINDS
       detected at GALAHAD_KINDS_SINGLE@<end-of-statement>
Error: ../src/external/mkl_pardiso/mkl_pardisot.F90, line 5: Redeclaration of symbol RPC_ from USEd module GALAHAD_KINDS
       detected at GALAHAD_KINDS_SINGLE@<end-of-statement>
```

dpo:
@nimgould The `#include` bug also occurs with the Intel compiler. You can see it happen on the GitHub Actions VMs too: https://github.com/ralna/GALAHAD/actions/runs/4451917184/jobs/7819096974#step:9:60

.github/workflows/standalone.yml
```diff
@@ -7,87 +7,89 @@ on:
...
-          .github/scripts/${{ matrix.script }} "${{ matrix.hpckit }}" all
-      - name: Setup GALAHAD build
+        uses: modflowpy/install-intelfortran-action@v1
+
+      - name: Setup GALAHAD
         shell: bash
         run: |
-          [[ "${{ matrix.compiler }}" == "intel" ]] && source /opt/intel/oneapi/setvars.sh
-          meson setup builddir -Dciface=true
+          meson setup builddir
```
amontoison, line 71:
```suggestion
          meson setup builddir -Dssids=false
```

amontoison: If we want to test GALAHAD without `ssids`.

amontoison:
> Compilation then fails in `src/pardiso/mkl_pardisot.F90`, where the compiler complains that variables from `GALAHAD_KINDS` are being redefined.

With preprocessing we have two `GALAHAD_KINDS_single` or `GALAHAD_KINDS_double` [here](https://github.com/ralna/GALAHAD/blob/galahad5/src/kinds/kinds.F90#L58-L98).
Some compilers probably don't allow that.

amontoison:
> Hi guys. I just tried to compile GALAHAD from this branch with clang/clang++/nagfor. On macOS, I have to use the command
> 
> ```shell
> CC=clang CXX=clang++ FC=nagfor FFLAGS=-fpp meson setup builddir_nag
> ```
> 
> because `nagfor` does not preprocess by default (so it returns an error at the first `#ifdef`).

For JuliaHSL, I added
```
if fc.get_id() == 'gcc'
	add_global_arguments('-cpp', language : 'fortran')
elif fc.get_id() == 'intel' or fc.get_id() == 'nagfor'
	add_global_arguments('-fpp', language : 'fortran')
elif fc.get_id() == 'flang'
	add_global_arguments('-c', language : 'fortran')
endif
```
to always preprocess.

dpo:
The `rpdt` issue is resolved. I also removed the `version` file. Interestingly, the Intel compiler doesn't complain about GALAHAD_KINDS on the virtual machines... The remaining failures are all on Windows.

nimgould:
Super, thanks. The version file had nothing to do with the C preprocessor, that is simply a name coincidence. It was simply an indication of which version of GALAHAD was in the latest zip/tar file, so that users could report. I guess this will all be superseded when the new release goes live.

I still don't follow why the kinds.F90 file causes issues. Quite deliberately there are no preprocessor commands in this file, and so  GALAHAD_KINDS_single, GALAHAD_KINDS_double and GALAHAD_KINDS_precision are three separate modules.

I'm glad that the rpdt issue has been settled.

dpo:
Sorry if this is a silly question but
```
❯ find builddir_nag -name "*galahad_kinds*"                                                                 18:08:50
builddir_nag/libspmf_double.dylib.p/galahad_kinds.mod
builddir_nag/libspmf_double.dylib.p/galahad_kinds_single.mod
builddir_nag/libspmf_double.dylib.p/galahad_kinds_double.mod
builddir_nag/libspmf_double.dylib.p/galahad_kinds_precision.mod
builddir_nag/libspmf_single.dylib.p/galahad_kinds.mod
builddir_nag/libspmf_single.dylib.p/galahad_kinds_single.mod
builddir_nag/libspmf_single.dylib.p/galahad_kinds_double.mod
builddir_nag/libspmf_single.dylib.p/galahad_kinds_precision.mod
builddir_nag/libpardiso.dylib.p/galahad_kinds.mod
builddir_nag/libpardiso.dylib.p/galahad_kinds_single.mod
builddir_nag/libpardiso.dylib.p/galahad_kinds_double.mod
builddir_nag/libpardiso.dylib.p/galahad_kinds_precision.mod
# ...
```
and the list continues with many other libraires. Is it expected that the `GALAHAD_KINDS` module should be part of every library out there? I would have expected it to be part of `libgalahad` and that dependents would find it when linking against `libgalahad`.

amontoison:
> Sorry if this is a silly question but
> 
> ```
> ❯ find builddir_nag -name "*galahad_kinds*"                                                                 18:08:50
> builddir_nag/libspmf_double.dylib.p/galahad_kinds.mod
> builddir_nag/libspmf_double.dylib.p/galahad_kinds_single.mod
> builddir_nag/libspmf_double.dylib.p/galahad_kinds_double.mod
> builddir_nag/libspmf_double.dylib.p/galahad_kinds_precision.mod
> builddir_nag/libspmf_single.dylib.p/galahad_kinds.mod
> builddir_nag/libspmf_single.dylib.p/galahad_kinds_single.mod
> builddir_nag/libspmf_single.dylib.p/galahad_kinds_double.mod
> builddir_nag/libspmf_single.dylib.p/galahad_kinds_precision.mod
> builddir_nag/libpardiso.dylib.p/galahad_kinds.mod
> builddir_nag/libpardiso.dylib.p/galahad_kinds_single.mod
> builddir_nag/libpardiso.dylib.p/galahad_kinds_double.mod
> builddir_nag/libpardiso.dylib.p/galahad_kinds_precision.mod
> # ...
> ```
> 
> and the list continues with many other libraires. Is it expected that the `GALAHAD_KINDS` module should be part of every library out there? I would have expected it to be part of `libgalahad` and that dependents would find it when linking against `libgalahad`.

If we have multiple libraries, yes it's expected because we don't a specific library for `GALAHAD_KINDS`. But we decided to merge all libraries together to just have a `libgalahad.so`. I did it in the branch `meson_dependencies`, I will see if I can merge it.

amontoison:
@nimgould We have 4 four errors in the tests. They are all related to the C interface of `dps` (`dpst` and `dpstf`).

dpo:
Now the build fails on macOS M1 will all compiler combinations :-(

Better setup CI on M1 using CirrusCI.

nimgould:
Very odd. dps tests in C work here with all compilers. Any chance of getting more details?

dpo:
@nimgould The GLTRT crash seems to simply be due to an array overrun. `data%RMINVRS` is allocated to have size `data%itmax + 1` (and apparently, `data%itmax = 100`). I changed `gltr.F90` as follows:
```diff
diff --git a/src/gltr/gltr.F90 b/src/gltr/gltr.F90
index 4f5d2bbf..c4263945 100644
--- a/src/gltr/gltr.F90
+++ b/src/gltr/gltr.F90
@@ -1532,6 +1532,9 @@

 !  Update the solution estimate

+      WRITE(6, *) 'len RMINVRS = ', data%itmax + 1
+      WRITE(6, *) 'itp1 = ', itp1
+      WRITE(6, *) 'dim_sub = ', data%dim_sub
       data%rminvr = data%RMINVRS( itp1 )
       IF ( data%iter /= 0 ) THEN
         X = X + data%tau
```
and it's clear that `itp1` becomes much larger than `data%itmax + 1`:
```
 len RMINVRS =          101
 itp1 =       254757
 dim_sub =            0
```
The test
```fortran
IF ( data%iter >= data%itmax ) GO TO 910
```
should probably occur independently of whether `itp1 == data%dim_sub` or not.

Of course, this isn't the real cause of the issue, but at least, it avoids a segfault. After that, `gltrt_single` seems to be in an infinite loop. It doesn't even time out. So perhaps another stopping test is missing?!

dpo:
`glrtt_single` is also in an infinite loop after seemingly hitting the hard case. Here is the output I get: https://gist.github.com/dpo/84bacdcffe675ee122a83fd06e03136a

dpo:
I can't reproduce the error with the Intel compiler on macOS locally. My build fails because of the KINDS module. I'm still not sure I understand how meson knows that that module is a dependency.

dpo:
@nimgould I added a section to `standalone.yml` to allow ssh access to the GitHub Actions virtual machine. See my commit above (https://github.com/ralna/GALAHAD/pull/85/commits/a0e1b5e5b2cff0262d23700f5839d82d1ae78ccb). Simply uncomment the section and commit the change to this branch. Push, and the workflows will run with debugging enabled. Keep an eye on the output of the workflow you're interested in (say, macos/intel). Before the build step, you should see lines of the form `ssh blablabla.tmate.io`. Copy-paste the ssh command into your terminal, and you will connect to the VM. There, you can run the same commands you would in your local terminal.

It's not super convenient, but until the `standalone.yml` workflow is committed to the `master` branch, it'll have to do.

nimgould:
I am looking at the GLTR crash above, and can't reproduce it here, nor would my understanding of the 25 year old logic suggest that it could happen; data%iter should never be larger than control%itmax. The value itp1 is data%iter + 1 which should be at most 101. Thus all I can believe is that there is a memory overwrite onto data%iter somewhere. That'll be tough to track, although I suppose that gdb might be our shepherd.

dpo:
@nimgould No luck with gdb so far:
```
(gdb) run
Starting program: /Users/dpo/dev/ralna/GALAHAD/builddir_gnu_intel/gltrt_single
[New Thread 0x1e03 of process 42203]
[New Thread 0x2103 of process 42203]
During startup program terminated with signal SIGTRAP, Trace/breakpoint trap.
(gdb) bt
No stack.
```

Valgrind no longer runs on Macs, unfortunately.

If I move the test on `data%iter` outside the test on `dim_sub`
```diff
@@ -1547,6 +1550,7 @@

 !  If the approximate minimizer is complete, exit

+        IF ( data%iter >= data%itmax ) GO TO 910
       IF ( itp1 == data%dim_sub ) THEN
 !     IF ( itp1 > data%dim_sub ) THEN
         inform%iter_pass2 = data%iter
diff --git a/src/gltr/gltrt.F90 b/src/gltr/gltrt.F90
index a6b7e8bf..58ede644 100644
--- a/src/gltr/gltrt.F90
+++ b/src/gltr/gltrt.F90
```
then the iterations stop at iteration 101, but the program still hangs as it tries to terminate. Investigating.

dpo:
We're stuck in this part of `GLTR_leftmost_eigenvalue` (line 2472):
```fortran
      ITER_LOOP : DO
        WRITE(6, *) 'main loop with iter = ', iter
        iter = iter + 1

!  Compute the inertia of T - e I by implicitly factoring the matrix

        neg = 0
        piv = D( 1 ) - e
        dpiv = - one

!  If a zero pivot is encountered, reset the upper bound

        IF ( piv == zero ) THEN
          WRITE(6, *) 'zero pivot'
          u = e
          e = half * ( l + u )
          CYCLE ITER_LOOP
```
I observe the following output:
```
 main loop with iter =      4194104
 zero pivot
 main loop with iter =      4194105
 zero pivot
 main loop with iter =      4194106
 zero pivot
```
and so on, without end.

I notice that there remain constants that are not defined in a precision-dependent manner. I think it's capital that we hunt them all down and redefine them properly. For example:

```diff
diff --git a/src/gltr/gltr.F90 b/src/gltr/gltr.F90
index 4f5d2bbf..06c61ca0 100644
--- a/src/gltr/gltr.F90
+++ b/src/gltr/gltr.F90
@@ -78,7 +78,7 @@
       REAL ( KIND = rp_ ), PARAMETER :: one = 1.0_rp_
       REAL ( KIND = rp_ ), PARAMETER :: two = 2.0_rp_
       REAL ( KIND = rp_ ), PARAMETER :: ten = 10.0_rp_
-      REAL ( KIND = rp_ ), PARAMETER :: roots_tol = ten ** ( - 12 )
+      REAL ( KIND = rp_ ), PARAMETER :: roots_tol = 100 * EPSILON( ONE )
```
and
```diff
@@ -2423,13 +2427,16 @@
       INTEGER ( KIND = ip_ ) :: i, neg, nroots
       REAL ( KIND = rp_ ) :: l, u, e, piv, dpiv, e_trial, infinity
       REAL ( KIND = rp_ ) :: root1, root2, tol_interval, sum, prod
-      REAL ( KIND = rp_ ), PARAMETER :: pert = ten ** ( - 6 )
+      REAL ( KIND = rp_ ), PARAMETER :: pert = EPSILON( ONE ) ** 0.33
```

amontoison:
@nimgould It seems that we have errors with other `inc*.F90`, can you fix them when you have time?
I will try to understand why the `GALAHAD_KIND_PRECISION` modules are not generated for Windows after with the cross-compiler.

nimgould:
I don't know what is happening here. All is fine with the makefile build and test here. However, the meson/ninja version reports (for example) 
/usr/bin/ld: libgalahad_single.so.p/src_qpa_inqpa.F90.o: in function `main':
427
/home/runner/work/GALAHAD/GALAHAD/builddir/../src/qpa/inqpa.F90:23: multiple definition of `main'; 
Are you trying to put the in* binaries into the library? They shouldn't be, as they are main programs

nimgould:
I am worried by  anything that says
main loop with iter =      4194104
how on earth does iter get so big?
Once again, this suggests to me a memory overwrite/corruption

dpo:
> Agreed that there should be different tolerances for the single and double versions. My main worry is not to change the double ones, as these have been relied on for years, and I don't want to upset existing users, and packages that depend on them. But, yes, a good opportunity for the single precision version. The usual word of caution, though, setting a stopping rule on the gradient for (say) 0.001 may be all that is warranted, but this does mean that we may stop far from a minimizer, and we need to make sure that users appreciate the limitations of single precision.

This version of GALAHAD will be so different that it would be fair to tell users that some double precision tolerances/data have changed slightly to accommodate multi precision support. I think that when someone uses LAPACK in single precision, they don't expect a backward error of 1.0e-12. A gradient of 0.001 in single is sqrt(eps), and pretty much the best one can hope for.

Would you be able to run Valgrind on gltrt locally (on Linux) in single precision to see if anything turns up? I found a fork of Valgrind for macOS but it doesn't support arm64. I'll have to try on an old Intel Mac. I haven't been able to make any other leak detection tool work. I'm trying compiler sanitizers, but again, the gcc/gfortran/arm64 combination doesn't seem to be supported.

nimgould:
I will have a go with valgrind. I did valgrind everything a while ago, and fixed a load of memory leaks, but it is possible that we now have more.

I'm not so worried about tolerances at the outer (user) level, more for their internal use, for instance a linear solver at the heart of an optimization method that doesn't achieve the residual required. This has been honed carefully over the years in double, but not so in single.
And this probably relates to errors we encounter. I'm not saying that we should (indeed we must), but it may be a lengthy process

nimgould:
I tried valgrind on gltrt in single precision, but this is fine -
   total heap usage: 412 allocs, 412 frees, 103,162 bytes allocated
  ERROR SUMMARY: 0 errors from 0 contexts

With these massive iter values, can you put in a write every time iter is changed, so that we can at least see that this isn't simply a memory overwrite?

dpo:
Thank you. I did put a write at each pass through the loop, but it never ends. The piece of code I quoted above results in an infinite loop in single precision. I'm not convinced the error is memory related yet; I'm thinking it's either a double-precision value being used somehow, an erroneous cast, or something of that sort.

We enter GLTR_leftmost_eigenvalue with this data:
```
 entering GLTR_leftmost_eigenvalue
  it neg        l               e               u              piv
   0   0  0.00000000E+00  0.00000000E+00  2.29588740E-39        -
```
which already looks highly suspicious.

I forced the loop in GLTR_leftmost_eigenvalue to stop after 100 passes. The loop is otherwise infinite because
```
 main loop with iter =           98
 zero pivot: u =    0.00000000     , e =    0.00000000
 main loop with iter =           99
 zero pivot: u =    0.00000000     , e =    0.00000000
 main loop with iter =          100
 iteration    101 leftmost eigenvalue =   0.00000000000000E+00
```
(`u` and `e` were 0 all along).

nimgould:
Thank you. 

For the makemaster version, I plan to use a different mechanism. When building, I will create 3 libraries libgalahad_metis4.a, libgalahad_metis5.a and libgalahad_dummy_metis.a that will contain the three interfaces (+ C file) that you provided. The GALAHAD run scripts will have a variable METIS which will be set by default to -ibgalahad_dummy_metis (since we know that this will always work even if no metis is available). Users can either change the architecture file, or set the environment variable METISLIB to override this. I will also replace explicit mentions of metis_nodend by galahad_metis in the other galahad routines; I will still have to use the -Dmetis_nodend=galahad_metis flag when compiling the real HSL packages, if any. The only minor thing is that we will then need to include the metis.h file that is provided when metis5 is present; I'll copy this into GALAHAD's include directory.

Any comments? I realise that the meson version is different, but the makemaster version is intended to work without knowing anything about the existence of external packages.

amontoison:
It's a good idea to keep a copy of metis.h here, It will simplify the Meson build too because I added a variable `libinclude_include` to find this header file.

I like your proposition with the 3 libraries libgalahad_metis_*.a. I can't do that because I merged everything in libgalahad.so and I can only have one `galahad_metis` library.

I only have one comment about preprocessing, you will also need `-DMETIS_NODEND=galahad_metis` for some HSL packages like MA57. Preprocessing flags are sensible to capital letters.

nimgould:
Good point, I will add that

amontoison:
Nick, only test is failing on Windows. It's `qpt` in single precision (timeout). Do you have an idea of the problem with the logs?

src/spral/metis4_wrapper.F90
```diff
@@ -156,12 +156,12 @@ subroutine metis_order64(n,ptr,row,perm,invp,flag,stat)
 
     ! Carry out ordering
     metis_opts(1) = 0 ! MeTiS defaults
-    call metis_nodend(n,ptr2,row2,1,metis_opts,invp,perm)
```
amontoison, line 159: Nick, don't you have an infinite `loop`if `galahad_metis` calls `galahad_metis`?

nimgould: Do I? I don't see it

amontoison: I checked again and you call `metis_nodend` on the last commit of `multi-precision` so you already fixed and It was just an old commit. I resolve the conversation 🙂

amontoison:
@nimgould 
We have some errors when we compile the C interface of `llst` .

nimgould:
Ooops, yes, the single precision preprocessor tries to do 

#define llst_import llst_import_s
#define llst_import_s llst_import_s_s

but we should instead do

#define llst_import_s llst_import_s_s
#define llst_import llst_import_s

(there is unfortunately a procedure called llst_import_s and this clashes
with adding a _s to the single versions). I will just test to see if this works

nimgould:
Unfortunately the preprocessor still gets in a muddle. I have changed the name of the function so that import_s is now import_scaling, and all is now ok ... I hope!

amontoison:
> Unfortunately the preprocessor still gets in a muddle. I have changed the name of the function so that import_s is now import_scaling, and all is now ok ... I hope!

You fixed the issue. Great!
Nick, on Windows platforms we can't have filename with basename `aux` like `aux.rst`.
`aux.*` files have a special application on Windows.

I renamed `aux.rst` into `auxiliairy.rst` last week but you push it again yesterday.

nimgould:
Sorry, I didn't see that push ... and it doesn't appear in my doc/Python directory. I'll rename it here as you suggested, and modify the other files that refer to it.

amontoison:
It's this commit: https://github.com/ralna/GALAHAD/pull/85/commits/15d1989b426d3fd2e8c1d7490832a06f2183390a

I will rebase the branch `python_interface` to test the compilation of the Python interface of `llst`.

nimgould:
How odd, it didn't seem to arrive here. Possibly because of one of my periodic git breakages.

At some stage, we also need to think about how/if meson can cope with the matlab interfaces. These are a bit tricky as they require specific versions of the compilers

amontoison:
> At some stage, we also need to think about how/if meson can cope with the matlab interfaces. These are a bit tricky as they require specific versions of the compilers

I propose to work on it after the support of Python interfaces with Meson.

PR #93, opened 2023-03-20
[Meson] Update how we compile files for the CUTEST interface

amontoison:
@nimgould @dpo 
I found that the file `kinds/kinds.F90` was compiled two times. The issue with the dependency tree / compilation order could be related.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -235,12 +235,9 @@ subdir('src/ssids')
 subdir('src/spral')
 
 if build_single
-  sources_single = [libgalahad_single_src, libgalahad_integer_src, libgalahad_src]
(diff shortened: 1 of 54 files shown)


amontoison:
@nimgould I have the following error:
```shell
../src/ccqp/inccqp.F90:837:80:

  837 |         CALL CCQP_solve( prob, data, CCQP_control, CCQP_inform, C_stat, B_stat )
      |                                                                                1
Error: More actual than formal arguments in procedure call at (1)
```

nimgould:
Fixed. Will commit later. Thanks

amontoison:
Nick, it was a pull request, not an issue. I still want to merge my modifications after your fix ;)

PR #94, opened 2023-03-21
[Meson] AMPL interface

amontoison:
@nimgould Do we also need `libcutest` if we want to compile the files in the `src/ampl` folder?

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -88,6 +88,11 @@ option('libmkl_pardiso',
        value : 'mkl_intel_lp64',
        description : 'option for Intel MKL PARDISO library switching')
 
+option('libampl',
+       type : 'string',
+       value : 'ampl',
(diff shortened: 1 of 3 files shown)


nimgould:
Let's not worry about the ampl interfaces, they are very old and I am not sure I now
know how they work. @dpo is the magician here

meson_options.txt
```diff
@@ -88,6 +88,11 @@ option('libmkl_pardiso',
        value : 'mkl_intel_lp64',
        description : 'option for Intel MKL PARDISO library switching')
 
+option('libampl',
+       type : 'string',
+       value : 'ampl',
+       description : 'option for Intel AMPL library switching')
```
amontoison, line 94:
```suggestion
       description : 'option for AMPL library switching')
```

PR #98, opened 2023-03-28
[Meson] Update how we compile examples

amontoison:
- Fortran examples
- C examples


amontoison:
I will do the same thing with the tests.

PR #103, opened 2023-03-29
[Meson] the modules appear in alphabetical order

amontoison:


amontoison:
Ubuntu build failed so it means that the dependency tree generated by Meson is wrong.
We must preprocess all files before the compilation.

amontoison:
@nimgould Is it possible to rename the following files?
- src/lancelot/extend.F90 -> src/lancelot/lancelot_extend.F90 
- src/spral/kinds.F90 -> src/spral/spral_kinds.F90
- src/spral/scaling.F90 -> src/spral/spral_scaling.F90

Meson preprocesses all files in the same folder and we a need different name for all files.
I don't want to break your build system.

nimgould:
Yes, I will do that later today. No problem. I'll let you know when I've fixed and tested this locally.

amontoison:
Thanks Nick!

PR #107, opened 2023-03-30
[Meson] Update galahad_headers

### LINKED ISSUE #104: Merge folders with header files?
@nimgould We have header files in `include` and `src/dum/include`. I propose to remove the `src/dum/include` folder.
We also have two header files in `src/ampl`. It will be easier to maintain if we only have one folder.

amontoison:
close #104 
Nick, can you review the PR?

### DIFF
--- a/include/meson.build
+++ b/include/meson.build
@@ -1,15 +1,43 @@
-galahad_headers = files('galahad_arc.h', 'galahad_bgo.h', 'galahad_blls.h', 'galahad_bqpb.h',
-                        'galahad_bqp.h', 'galahad_bsc.h', 'galahad_ccqp.h', 'galahad_cfunctions.h',
-                        'galahad_clls.h', 'galahad_convert.h', 'galahad_cqp.h', 'galahad_cro.h',
-                        'galahad_dgo.h', 'galahad_double.h', 'galahad_dps.h', 'galahad_dqp.h',
-                        'galahad_eqp.h', 'galahad_fdc.h', 'galahad_fit.h', 'galahad_glrt.h',
-                        'galahad_gls.h', 'galahad_gltr.h', 'galahad.h', 'galahad_hash.h',
-                        'galahad_icfs.h', 'galahad_ir.h', 'galahad_l2rt.h', 'galahad_lhs.h',
-                        'galahad_lms.h', 'galahad_lpa.h', 'galahad_lpb.h', 'galahad_lsqp.h',
-                        'galahad_lsrt.h', 'galahad_lstr.h', 'galahad_modules.h', 'galahad_nls.h',
-                        'galahad_precision.h', 'galahad_presolve.h', 'galahad_psls.h', 'galahad_python.h',
-                        'galahad_qpa.h', 'galahad_qpb.h', 'galahad_roots.h', 'galahad_rpd.h',
-                        'galahad_rqs.h', 'galahad_sbls.h', 'galahad_scu.h', 'galahad_sec.h',
-                        'galahad_sha.h', 'galahad_sils.h', 'galahad_single.h', 'galahad_slls.h',
-                        'galahad_sls_defaults.h', 'galahad_sls.h', 'galahad_trb.h', 'galahad_trs.h',
-                        'galahad_tru.h', 'galahad_ugo.h', 'galahad_uls.h', 'galahad_wcp.h')
+galahad_headers += files('spral_config.h', 'spral_procedures.h', 'spral_ssids.h',
+                         'spral_compat.hxx', 'spral_guess_topology.hxx', 'spral_hwloc_wrapper.hxx', 'spral_omp.hxx')
+
+if build_ciface
+    galahad_headers += files('galahad_arc.h', 'galahad_bgo.h', 'galahad_blls.h', 'galahad_bqpb.h',
+                             'galahad_bqp.h', 'galahad_bsc.h', 'galahad_ccqp.h', 'galahad_cfunctions.h',
+                             'galahad_clls.h', 'galahad_convert.h', 'galahad_cqp.h', 'galahad_cro.h',
+                             'galahad_dgo.h', 'galahad_double.h', 'galahad_dps.h', 'galahad_dqp.h',
+                             'galahad_eqp.h', 'galahad_fdc.h', 'galahad_fit.h', 'galahad_glrt.h',
+                             'galahad_gls.h', 'galahad_gltr.h', 'galahad.h', 'galahad_hash.h',
+                             'galahad_icfs.h', 'galahad_ir.h', 'galahad_l2rt.h', 'galahad_lhs.h',
+                             'galahad_lms.h', 'galahad_lpa.h', 'galahad_lpb.h', 'galahad_lsqp.h',
+                             'galahad_lsrt.h', 'galahad_lstr.h', 'galahad_modules.h', 'galahad_nls.h',
+                             'galahad_precision.h', 'galahad_presolve.h', 'galahad_psls.h',
+                             'galahad_qpa.h', 'galahad_qpb.h', 'galahad_roots.h', 'galahad_rpd.h',
+                             'galahad_rqs.h', 'galahad_sbls.h', 'galahad_scu.h', 'galahad_sec.h',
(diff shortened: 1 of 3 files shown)


dpo:
@amontoison When you open a PR, please include a description of what problem it solves and how so we can have an idea without having to comb through the diff. Thanks.

amontoison:
> I think you need to explain this to me in more detail. In particular, I don't want it to break the makefile build at the moment, I still have to rely on this for development. Where will the external headers lie on the distributed tree? They can't all lie in ./include as some users may not have access to the external packages, but some might (I am thinking, as always, as the install being for many users on a single system). If they are all in ./include, there may be clashes. Do they still occur in dum/include and then copied?

Sorry @nimgould, I should add a more detailed description. This PR will not break the makefile build.
I don't modify the include files in the GALAHAD source tree. The Meson variable `galahad_headers` is just a list of all headers that we want to install with Meson.
It means that all headers in the array `galahad_headers` will be installed in `$prefix/include`.
`$prefix` is the folder where GALAHAD will be installed.

include/meson.build
```diff
@@ -1,15 +1,43 @@
...
-                        'galahad_icfs.h', 'galahad_ir.h', 'galahad_l2rt.h', 'galahad_lhs.h',
-                        'galahad_lms.h', 'galahad_lpa.h', 'galahad_lpb.h', 'galahad_lsqp.h',
-                        'galahad_lsrt.h', 'galahad_lstr.h', 'galahad_modules.h', 'galahad_nls.h',
-                        'galahad_precision.h', 'galahad_presolve.h', 'galahad_psls.h', 'galahad_python.h',
-                        'galahad_qpa.h', 'galahad_qpb.h', 'galahad_roots.h', 'galahad_rpd.h',
-                        'galahad_rqs.h', 'galahad_sbls.h', 'galahad_scu.h', 'galahad_sec.h',
-                        'galahad_sha.h', 'galahad_sils.h', 'galahad_single.h', 'galahad_slls.h',
-                        'galahad_sls_defaults.h', 'galahad_sls.h', 'galahad_trb.h', 'galahad_trs.h',
-                        'galahad_tru.h', 'galahad_ugo.h', 'galahad_uls.h', 'galahad_wcp.h')
+galahad_headers += files('spral_config.h', 'spral_procedures.h', 'spral_ssids.h',
```
amontoison, line 1: Headers that we will install in all cases.

include/meson.build
```diff
@@ -1,15 +1,43 @@
...
+                             'galahad_icfs.h', 'galahad_ir.h', 'galahad_l2rt.h', 'galahad_lhs.h',
+                             'galahad_lms.h', 'galahad_lpa.h', 'galahad_lpb.h', 'galahad_lsqp.h',
+                             'galahad_lsrt.h', 'galahad_lstr.h', 'galahad_modules.h', 'galahad_nls.h',
+                             'galahad_precision.h', 'galahad_presolve.h', 'galahad_psls.h',
+                             'galahad_qpa.h', 'galahad_qpb.h', 'galahad_roots.h', 'galahad_rpd.h',
+                             'galahad_rqs.h', 'galahad_sbls.h', 'galahad_scu.h', 'galahad_sec.h',
+                             'galahad_sha.h', 'galahad_sils.h', 'galahad_single.h', 'galahad_slls.h',
+                             'galahad_sls_defaults.h', 'galahad_sls.h', 'galahad_trb.h', 'galahad_trs.h',
+                             'galahad_tru.h', 'galahad_ugo.h', 'galahad_uls.h', 'galahad_wcp.h')
+endif
```
amontoison, line 20: Headers that we will install if we compile the C interface.

include/meson.build
```diff
@@ -1,15 +1,43 @@
...
+
+if build_pythoniface
+  galahad_headers += files('galahad_python.h')
+endif
```
amontoison, line 24: Headers that we will install if we compile the Python interface.

include/meson.build
```diff
@@ -1,15 +1,43 @@
...
+                           'ssids_cpu_NumericSubtree.hxx', 'ssids_cpu_SmallLeafNumericSubtree.hxx', 'ssids_cpu_SmallLeafSymbolicSubtree.hxx',
+                           'ssids_cpu_SymbolicNode.hxx', 'ssids_cpu_SymbolicSubtree.hxx', 'ssids_cpu_ThreadStats.hxx',
+                           'ssids_cpu_Workspace.hxx', 'ssids_profile.hxx')
+endif
+
+if not libhsl.found()
+  galahad_headers += files('hsl_ma48.h', 'hsl_ma57.h', 'hsl_ma77.h', 'hsl_ma86.h',
+                           'hsl_ma87.h', 'hsl_ma97.h', 'hsl_mc64.h', 'hsl_mc68.h',
+                           'hsl_mi20.h', 'hsl_mi28.h')
+endif
```
amontoison, line 43: Dummy headers that we will install only if we didn't find JuliaHSL.

include/meson.build
```diff
@@ -1,15 +1,43 @@
...
+                           'ssids_cpu_AppendAlloc.hxx', 'ssids_cpu_BlockPool.hxx', 'ssids_cpu_BuddyAllocator.hxx',
+                           'ssids_cpu_cpu_iface.hxx', 'ssids_cpu_factor.hxx', 'ssids_cpu_kernels_assemble.hxx',
+                           'ssids_cpu_kernels_block_ldlt.hxx', 'ssids_cpu_kernels_calc_ld.hxx', 'ssids_cpu_kernels_cholesky.hxx',
+                           'ssids_cpu_kernels_common.hxx', 'ssids_cpu_kernels_ldlt_app.hxx', 'ssids_cpu_kernels_ldlt_app.hxx',
+                           'ssids_cpu_kernels_ldlt_nopiv.hxx', 'ssids_cpu_kernels_ldlt_tpp.hxx', 'ssids_cpu_kernels_SimdVec.hxx',
+                           'ssids_cpu_kernels_verify.hxx', 'ssids_cpu_kernels_wrappers.hxx', 'ssids_cpu_NumericNode.hxx',
+                           'ssids_cpu_NumericSubtree.hxx', 'ssids_cpu_SmallLeafNumericSubtree.hxx', 'ssids_cpu_SmallLeafSymbolicSubtree.hxx',
+                           'ssids_cpu_SymbolicNode.hxx', 'ssids_cpu_SymbolicSubtree.hxx', 'ssids_cpu_ThreadStats.hxx',
+                           'ssids_cpu_Workspace.hxx', 'ssids_profile.hxx')
+endif
```
amontoison, line 37: Headers that we will install only if we only compile GALAHAD with `ssids`.

src/ampl/meson.build
```diff
@@ -2,4 +2,6 @@ if libampl.found()
   libgalahad_ampl_src += files('galahad_ampl.c', 'wrap_bqp.F90', 'wrap_cqp.F90',
                                'wrap_filtrane.F90', 'wrap_lancelot.F90', 'wrap_presolve.F90',
                                'wrap_qp.F90', 'wrap_qpa.F90', 'wrap_qpb.F90', 'wrap_qpc.F90')
+
+  galahad_headers += files('amplinter.h', 'galahad_ampl.h')
```
amontoison, line 6: Headers that we will install only if we only compile GALAHAD with the `AMPL` interface.

src/dum/meson.build
```diff
@@ -62,10 +62,12 @@ endif
 
 if not libsmumps.found()
   libgalahad_single_src += files('mumps.F90')
+  galahad_headers += files('include/smumps_struc.h')
```
amontoison, line 65: Dummy header that we will install only if we didn't find the single precision version of `MUMPS`.

src/dum/meson.build
```diff
@@ -62,10 +62,12 @@ endif
...
 endif
 
 if not libdmumps.found()
   libgalahad_double_src += files('mumps.F90')
+  galahad_headers += files('include/dmumps_struc.h')
```
amontoison, line 70: Dummy header that we will install only if we didn't find the double precision version of `MUMPS`.

amontoison (commented):
@nimgould 
I added some comments.

src/dum/meson.build
nimgould: Thank you @amontoison , that is clear. I'll merge

PR #108, opened 2023-04-01
see if sanitizers work across the board

dpo:
Sanitizers are tools that can help diagnose memory, addressing, and other types of errors.
See https://github.com/google/sanitizers.

### DIFF
--- a/.github/workflows/standalone.yml
+++ b/.github/workflows/standalone.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Setup GALAHAD
         shell: bash
         run: |
-          meson setup builddir --prefix=$GITHUB_WORKSPACE/galahad -Dpythoniface=true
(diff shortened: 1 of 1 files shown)


dpo:
@nimgould In this pull request, I build GALAHAD with sanitizers enabled. They are alternatives to Valgrind to diagnose various types of issues. We see issues on both Linux and macOS. They are not very verbose, but perhaps you will see more output if you try locally. The only difference is that you set up the build dir with the command
```console
meson setup -Db_sanitize=address,undefined builddir
```
then you run the unit tests normally, all at once or one at a time.

On my ARM64 Mac, one of those sanitizers suggested a possible cause for the `gltrt` failure in single precision. After investigating, I realized that `GLTR_ttrs` was called with `n = 0`. That is because `dim_sub = 0` earlier, but I haven't yet determined the cause of that. I suspect integer size mismatches (especially when we call BLAS/LAPACK), but it seems there may also be buffer overruns/leaks/etc.

amontoison:
Thanks @dpo!
We added it on `master`.

PR #111, opened 2023-04-13
Preprocess files before the compilation

amontoison:
We need a fix in Meson before that we can merge this PR.
I created an issue [here](https://github.com/mesonbuild/meson/issues/11504) with a small project to reproduce the error.

I hope that they will be able to solve it in a future release :crossed_fingers:. 
It will solve all our issues and we will be able to merge `libgalahad_single.${dlext}` and `libgalahad_double.${dlext}` together.


amontoison:
I found a workaround this morning to use `generator` instead of `fc.preprocess` but 10 minutes after that I opened #118 they found a way to fix it in Meson! (https://github.com/mesonbuild/meson/pull/11746)
It will avoid to adapt the `generator` commands that I added in #118 for each Fortran compiler.

amontoison:
The PR is merged, we will be able to use `fc.preprocess` with Meson 1.1.1.

PR #118, opened 2023-04-27
Use generators to preprocess Fortran files in single and double precision

amontoison:
@dpo @jfowkes  @nimgould 
I finally found how to just preprocess the Fortran files to generate the source files for single and double precision. :muscle: 
I probably need to adapt it for different Fortran compilers but it should fix all our issues.


amontoison:
I found the culprit... In some cases `kinds.F90` was not compiled because in `src/lapack/meson.build` we used `libgalahad_src = ...` instead of `libgalahad_src += ...`.
Maybe preprocessing all files before the compilation is not needed.

amontoison:
All Fortran and C tests based on the double precision version of GALAHAD passed on all platforms except (Windows with Intel compilers (the odd duck...).

I merge the PR and reactivate the tests in single precision such that we can improve the support of single precision in the future.

nimgould:
Thanks for all of this, Alexis. I'll get back to 5.0 once I get home and unpacked.

amontoison:
Enjoy your vacation Nick :wine_glass: 
I just wanted to finish this before May such that I can spend more time on writing my thesis.

PR #120, opened 2023-04-27
Try to preprocess and compile in the same time

amontoison:


amontoison:
It's not working, we must preprocess the source files before the compilation.

PR #124, opened 2023-05-16
Support METIS 5 in GALAHAD

amontoison:
In brief, I use fortran preprocessing to replace `metis_nodend` by `galahad_metis`.
We have three versions of `galahad_metis` (dummy, metis4, metis5).
They are based on the API of METIS 4 so we don't need to update the source code of GALAHAD.

I removed `src/dum/metis4.f` because it's replaced by `src/external/metis/dummy_metis.f90`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -131,7 +131,7 @@ galahad_c_tests = []
 # Headers and Fortran modules
 libgalahad_include = [include_directories('include'),
                       include_directories('src/dum/include'),
-                      include_directories('src/ampl')] + libhsl_modules
+                      include_directories('src/ampl')] + libmetis_include + libhsl_modules
 
 # TODO: -DSPRAL_NO_SCHED_GETCPU | DSPRAL_HAVE_SCHED_GETCPU
 extra_args = ['-DSPRAL_NO_SCHED_GETCPU']
--- a/meson.build
+++ b/meson.build
@@ -74,6 +74,9 @@ libpastix_path = get_option('libpastix_path')
 libmkl_pardiso_path = get_option('libmkl_pardiso_path')
 libampl_path = get_option('libampl_path')
 
+libmetis_include = get_option('libmetis_include')
+libmetis_version = get_option('libmetis_version')
+
 libhsl_modules = get_option('libhsl_modules')
 
 # Dependencies
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -167,3 +167,14 @@ option('libhsl_modules',
        type : 'array',
        value : [],
        description : 'Additional directories to search HSL modules')
+
+option('libmetis_include',
+       type : 'array',
+       value : [],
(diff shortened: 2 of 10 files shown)


amontoison:
> That looks very good, but I need to think about how to use it from the makefile build. I could set an environment variable (metis_version?) and based on this builds a galahad_metis library containing the relevant juliahsl_metis (+ dependecies).

`METIS_VERSION` is a good variable name. You currently use the variable `METIS4` and `METIS5` in archdefs but I don't think that can coexist because they have the same soname (we link with -lmetis` in both cases).
So it makes more senses to have the two variables `METIS` and `METIS_VERSION`.

For the name of shared library generated by your build system `libgalahad_metis.so` is great.
For information, I merged it in `libgalahad.so` for the Meson build system.

> What I am not so sure about is what happens if the user provides their own binary hsl library (that they bought sometime in the past), rather than one that we can preprocess the name metis_nodend -> juliahsl_metis

Your `makemaster` build system can't link with an HSL library. I checked the [readme](https://raw.githubusercontent.com/ralna/GALAHAD/master/doc/README) and I understand that the user needs to copy the source code of the HSL packages that they bought in a specific folder and after it is compiled by GALAHAD.
Am I wrong?

nimgould:
OK, I'll see if I can sort out the metis mess in the makemaster system.

A user can easily change the name of the HSL (and any other provided) library.
That is what the $HSL variable is for, so if a user has a compiled library libhsl.a (or .so), setting HSL = -lhsl will do the job

nimgould:
If HSL is unset, the run scripts will default to -lgalahad_hsl. The README is simply for anyone who has followed the instructions to register for and download HSL , and wants the GALAHAD makefile install to build the HSL binary for them.

nimgould:
Actually, the GALAHAD run scripts do not do what is needed (they ignore the necessary shell variables), but it is easy to modify them so that they do. To avoid conflicts, we will require a name like HSLLIB so that if we mention ${HSLLIB--lgalahad_hsl} this will default to -lgalahad_hsl if HSSLIB is undefined but to a the value $HSLLIB if it is ... I just learned something new about bash!

nimgould:
OK, this is now live. For each external library "pack", the cutest and qplib run scripts (hal and dgal) now check (in order) for $PACKLIB then $PACK and finally default to -lgalahad_pack. The makemaster architecture files contain defaults for each PACK, but this is overridden by $PACKLIB if specified. Man pages also updated.

nimgould:
Could you put the external/metis directory in the muti-precision branch, using your usual git wizzardry, please. I have no idea how to ... except by cutting and pasting!!. Thank you

amontoison:
> Could you put the external/metis directory in the multi-precision branch, using your usual git wizzardry, please. I have no idea how to ... except by cutting and pasting!!. Thank you

I will check that the CI builds are not all failing and after I merge the PR. :mage:

PR #128, opened 2023-05-19
Use BLAS and LAPACK routines provided by GALAHAD for CI builds

amontoison:
It seems that some platforms find `libblas.${dlext}` and  `liblapack.${dlext}`.
It could explain why we have tests that fail only on some platforms.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -16,9 +16,9 @@ task:
     echo "GALAHAD=$CIRRUS_WORKING_DIR" >> $CIRRUS_ENV
   configure_script: |
     if [ "$(uname -s)" = "FreeBSD" ]; then
-      FC=gfortran12 CC=gcc12 CXX=g++12 meson setup builddir --prefix=~/galahad
+      FC=gfortran12 CC=gcc12 CXX=g++12 meson setup builddir --prefix=~/galahad -Dlibblas= -Dliblapack=
     else
-      FC=gfortran-12 CC=gcc-12 CXX=g++-12 meson setup builddir --prefix=~/galahad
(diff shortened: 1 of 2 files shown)


amontoison:
@dpo @jfowkes @nimgould 
The tests were failing on Mac because Meson automatically uses `libblas.dylib` and `liblapack.dylib` provided by the system. The issue is that these were not updated for 15 years before the release 13.3....
We only have one remaining test that is failing on Windows and after all tests (single / double) pass on all platforms for the GCC compiler.

amontoison:
I got these information for the failing test on Windows:  
```
================================== 156/373 ===================================
test:         GALAHAD:qp+single+fortran / qpt_single
start time:   07:02:18
duration:     30.08s
result:       exit status 1

 error exit tests 

 3: QP_solve exit status =     -3
 4: QP_solve+solver exit status =    -65    -4
18: QP_solve+solver exit status =    -65   -18

nimgould:
Yes, I saw this. The next test is supposed to set a CPU time limit of 0 seconds, and to return if this is exceeded. It seems not to ... so maybe I should increase this to 0.00001 seconds? The test succeeds on all other platforms.

amontoison:
I had a similar issue with Julia package Krylov.jl: https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/730#issuecomment-1538454646
The C functions that returns the elapsed time in Windows was only update every [milliseconds](https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/730#issuecomment-1538518662) and if you require less time to solve the problem, nothing is triggered.

PR #131, opened 2023-05-25
Update how we link with OpenMP for nagfor and ifort compilers

amontoison:


amontoison:
@nimgould 
This PR fixed the internal error with Intel compilers for Mac and Linux platforms.

PR #137, opened 2023-06-09
[Meson] Fix how we compile with OpenMP

amontoison:


amontoison:
@nimgould
Do we use OpenMP in some Fortran files of GALAHAD or only in the C++ code of SSIDS?

nimgould:
Yes, we use it in other places, e.g., in galahad's cqp module

amontoison:
Thanks Nick!

PR #138, opened 2023-06-10
Try the new BLAS and LAPACK on MacOS 13.4

amontoison:


amontoison:
CirrusCI uses Mac 13.1, which has the old BLAS and LAPACK symbols (< 13.3).
I can test if we have undefined symbols.

PR #139, opened 2023-06-13
Use OpenBLAS as the default library for BLAS and LAPACK routines

amontoison:
@nimgould 
I updated the default value for BLAS and LAPACK libraries.
It's `libopenblas.${dlext}` by default now.
If it's not found, we compile our own version.
  
I don't want that we link with the `libblas.dylib` and `liblapack.dylib` provided by Apple.


nimgould:
Sounds good. I'll continue to compile the source blas & Lapack for the makefile
version, as the intention there is to have something that is "guaranteed" to
work (and users can change the libraries linked after the fact).

PR #140, opened 2023-06-13
Comment #pragma omp cancellation point taskgroup in C++ code of SSIDS

amontoison:
@jfowkes @nimgould 
By commenting the lines with `#pragma omp cancellation point taskgroup` in SSIDS, I fixed the compilation with Intel compilers.

### DIFF
--- a/include/ssids_cpu_NumericSubtree.hxx
+++ b/include/ssids_cpu_NumericSubtree.hxx
@@ -114,7 +114,7 @@ public:
               #pragma omp atomic read
               my_abort = abort;
               if (!my_abort) {
-               #pragma omp cancellation point taskgroup
+               // #pragma omp cancellation point taskgroup
                try {
                   int this_thread = omp_get_thread_num();
 #ifdef PROFILE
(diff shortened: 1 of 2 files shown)


nimgould:
Great. This only comes up as a warning here, but I presume it is marked as an error in your tests. Since I have no idea what Jonathan Hogg had in mind, best to avoid the issue

nimgould:
The warnings have now gone, but as I said, the old code still worked here with the warnings

nimgould:
aside for this compiler failure, ifx is also not fully fortran-2017 compliant, and 
in particular doesn't support some of the standard iso C interface changes that
we need. Thus, for the time being, we should use ifort/icx not ifx/icx, the icx part
is mature.

jfowkes:
Although icx is still not available on macOS so there we must use ifort/icc (Intel still ships icc for macOS for now).

jfowkes:
Intel compiler download pages (select compilers on left menu):
https://www.intel.com/content/www/us/en/developer/articles/tool/oneapi-standalone-components.html

PR #146, opened 2023-06-22
[CI] Compile GALAHAD without ssids on Windows

amontoison:


amontoison:
@nimgould 
I updated CI builds and many tests are still failing.
The logs can be downloaded with: https://github.com/ralna/GALAHAD/suites/13799079677/artifacts/765272213

nimgould:
These most;ly seem to be related to external packages that are not available and to errors in calls to blas and lapack. This is beyond our control. There are other segfaults, but without a proper trace, I have no idea. I presume that there aren't options to windows to provide line numbers rather than Backtrace for this error:
#0  0x71e1e8ea
#1  0x71d94263
etc

I don't intend to let this hold back the release, but rather too mark this as an issue

PR #162, opened 2023-07-19
[Meson] Add Python interfaces

jfowkes:
Enable meson to build and test the GALAHAD Python interfaces. 

Also fix some minor bugs in the Python interfaces that were causing intermittent segfaults / test failures.

### DIFF
--- a/src/arc/meson.build
+++ b/src/arc/meson.build
@@ -5,7 +5,11 @@ if build_ciface
 endif
 
 if build_pythoniface
-  libgalahad_python_src += [['arc', files('Python/arc_pyiface.c')]]
+  libgalahad_python_src += [['arc', files('Python/arc_pyiface.c',
+                                          '../sls/Python/sls_pyiface.c','../ir/Python/ir_pyiface.c','../rqs/Python/rqs_pyiface.c',
+                                          '../glrt/Python/glrt_pyiface.c','../psls/Python/psls_pyiface.c','../dps/Python/dps_pyiface.c',
+                                          '../lms/Python/lms_pyiface.c','../sha/Python/sha_pyiface.c','../sils/Python/sils_pyiface.c',
+                                          '../rpd/Python/rpd_pyiface.c')]]
 endif
 
 if libcutest.found()
--- a/src/arc/meson.build
+++ b/src/arc/meson.build
@@ -1,7 +1,13 @@
 libgalahad_src += files('arc.F90')
+
 if build_ciface
   libgalahad_c_src += files('C/arc_ciface.F90')
 endif
+
+if build_pythoniface
+  libgalahad_python_src += [['arc', files('Python/arc_pyiface.c')]]
+endif
+
 if libcutest.found()
   libgalahad_cutest_src += files('usearc.F90', 'runarc_sif.F90')
 endif
@@ -11,6 +17,8 @@ galahad_tests += [['arc', 'arct', files('arct.F90')],
 
 galahad_c_tests += [['arc', 'arct_c', files('C/arct.c')],
                     ['arc', 'arctf_c', files('C/arctf.c')]]
+
+galahad_python_tests += [['arc', 'arc_py', files('Python/test_arc.py')]]
 
 galahad_examples += [['arcs', files('arcs.f90')],
                      ['arcs2', files('arcs2.f90')],
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -21,9 +21,9 @@ task:
     echo "OMP_PROC_BIND=TRUE" >> $CIRRUS_ENV
   configure_script: |
     if [ "$(uname -s)" = "FreeBSD" ]; then
-      FC=gfortran12 CC=gcc12 CXX=g++12 meson setup builddir --prefix=~/galahad -Dpythoniface=true
+      FC=gfortran12 CC=gcc12 CXX=g++12 meson setup builddir -Dpythoniface=true
     else
-      FC=gfortran-12 CC=gcc-12 CXX=g++-12 meson setup builddir --prefix=~/galahad -Dpythoniface=true
+      FC=gfortran-12 CC=gcc-12 CXX=g++-12 meson setup builddir -Dpythoniface=true
     fi
   build_script: |
     meson compile -C builddir
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -2,7 +2,7 @@ task:
   matrix:
     - name: FreeBSD
       freebsd_instance:
-        image: freebsd-13-1-release-amd64
+        image: freebsd-13-2-release-amd64
     - name: MacOS M1
       macos_instance:
         image: ghcr.io/cirruslabs/macos-monterey-base:latest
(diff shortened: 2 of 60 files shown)


jfowkes:
@amontoison for the windows tests it's not finding the Python modules, does PYTHONPATH / environment variables work differently on Windows? Log below:
```
MALLOC_PERTURB_=252 PYTHONPATH=D:\a\GALAHAD\GALAHAD/galahad\Lib\site-packages\ C:\hostedtoolcache\windows\Python\3.11.4\x64\python.exe D:\a\GALAHAD\GALAHAD\builddir\..\src/arc\Python/test_arc.py

Traceback (most recent call last):

  File "D:\a\GALAHAD\GALAHAD\builddir\..\src/arc\Python/test_arc.py", line 1, in <module>

    from galahad import arc

ImportError: DLL load failed while importing arc: The specified module could not be found.
```

jfowkes:
> We have a `arc\Python` instead of a `arc/Python` in the path.
> It could be the culprit.

That looks like a meson issue as it's happening for all the Python tests. Any ideas how to fix?

amontoison:
> > We have a `arc\Python` instead of a `arc/Python` in the path.
> > It could be the culprit.
> 
> That looks like a meson issue as it's happening for all the Python tests. Any ideas how to fix?

I need to update the `meson.build` because It's me that introduced the '\'.
https://github.com/ralna/GALAHAD/pull/162/files#diff-0ac3182fbcdd9d1e676dd2c57eaaca1ddb4db9ac7a565eaa4fa4869d0c76d3aeR25

src/arc/meson.build
```diff
@@ -12,6 +22,8 @@ galahad_tests += [['arc', 'arct', files('arct.F90')],
 galahad_c_tests += [['arc', 'arct_c', files('C/arct.c')],
                     ['arc', 'arctf_c', files('C/arctf.c')]]
 
+galahad_python_tests += [['arc', 'arc_py', files('Python/test_arc.py')]]
```
amontoison, line 25:
```suggestion
galahad_python_tests += [['arc', 'arc_py', files('Python\test_arc.py')]]
```

jfowkes:
@amontoison there's also an issue with the `/` in the PYTHONPATH and the exe call so it's probably not that:
```
PYTHONPATH=D:\a\GALAHAD\GALAHAD/galahad\Lib\site-packages\ 
D:\a\GALAHAD\GALAHAD\builddir\..\src/arc\Python/test_arc.py
```
as the test is actually found and running.

src/arc/meson.build
```diff
@@ -12,6 +22,8 @@ galahad_tests += [['arc', 'arct', files('arct.F90')],
 galahad_c_tests += [['arc', 'arct_c', files('C/arct.c')],
                     ['arc', 'arctf_c', files('C/arctf.c')]]
 
+galahad_python_tests += [['arc', 'arc_py', files('Python\test_arc.py')]]
```
amontoison, line 25:
```suggestion
galahad_python_tests += [['arc', 'arc_py', files('Python/test_arc.py')]]
```

jfowkes:
@amontoison we have identified the cause of the intermittent python test segfaults so all that is left is to figure out why the modules are not importing on Windows.

amontoison:
@jfowkes 

DLL dependencies for extension modules and DLLs loaded with [ctypes](https://docs.python.org/3.8/library/ctypes.html#module-ctypes) on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with [add_dll_directory()](https://docs.python.org/3.8/library/os.html#os.add_dll_directory) are searched for load-time dependencies. Specifically, PATH and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution. If your application relies on these mechanisms, you should check for [add_dll_directory()](https://docs.python.org/3.8/library/os.html#os.add_dll_directory) and if it exists, use it to add your DLLs directory while loading your library.

jfowkes:
> @jfowkes
> 
> DLL dependencies for extension modules and DLLs loaded with [ctypes](https://docs.python.org/3.8/library/ctypes.html#module-ctypes) on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with [add_dll_directory()](https://docs.python.org/3.8/library/os.html#os.add_dll_directory) are searched for load-time dependencies. Specifically, PATH and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution. If your application relies on these mechanisms, you should check for [add_dll_directory()](https://docs.python.org/3.8/library/os.html#os.add_dll_directory) and if it exists, use it to add your DLLs directory while loading your library.

@amontoison so what you're saying is that on Windows instead of PATH we have to use `os.add_dll_directory()` from within Python to point it to the site-packages directory under galahad where the PYD files are installed? I hate Windows 😠

jfowkes:
@amontoison @nimgould I have given up on making the Python interfaces run on Windows. The docs suggest that either:

1. The user copies the DLLs and PYDs to somewhere on the system path (presumably meaning C:\WINDOWS\system32) which is not an option (or even desirable if it was) for most Windows users
2. Each python script that requires DLLs/PYDs sets the paths before importing the relevant module, i.e. we would need to have something like:
```python
import sys
if sys.platform == 'win32':
    import os
    os.add_dll_directory('galahad/bin/')
    os.add_dll_directory('galahad/Lib/site-packages/')
    os.add_dll_directory('galahad/Lib/site-packages/galahad/')
```
where the specific paths depend on what prefix the user has chosen on install. This is also not desirable or usable.

I therefore propose that we disable the Python tests for Windows, no user is going to compile the python interfaces on Windows anyway since Windows users almost exclusively used precompiled Python collections such as Anaconda.

jfowkes:
@amontoison when you have time could you edit the yaml to switch off the python tests for windows? Once that's done we can finally merge this PR! I have rebased all the python fixes into a single commit to remove all the messy debug commits.

amontoison:
> @amontoison when you have time could you edit the yaml to switch off the python tests for windows? Once that's done we can finally merge this PR! I have rebased all the python fixes into a single commit to remove all the messy debug commits.

@jfowkes I updated the main `meson.build` to not run the Python tests if the platform is windows.

amontoison:
> @amontoison @nimgould I have given up on making the Python interfaces run on Windows. The docs suggest that either:
> 
>     1. The user copies the DLLs and PYDs to somewhere on the system path (presumably meaning C:\WINDOWS\system32) which is not an option (or even desirable if it was) for most Windows users
> 
>     2. Each python script that requires DLLs/PYDs sets the paths before importing the relevant module, i.e. we would need to have something like this _in every python script file_:
> 
> 
> ```python
> import sys
> if sys.platform == 'win32':
>     import os
>     os.add_dll_directory('galahad/bin/')
>     os.add_dll_directory('galahad/Lib/site-packages/')
>     os.add_dll_directory('galahad/Lib/site-packages/galahad/')
> ```
> 
> where the specific paths depend on what prefix the user has chosen on install. This is also not desirable or usable.
> 
> I therefore propose that we disable the Python tests for Windows, no user is going to compile the python interfaces on Windows anyway since Windows users almost exclusively use precompiled Python collections such as Anaconda.

Approved :+1: 
It's possible that everything works if `--prefix=...` is not to used but we can't easily test it.

PR #164, opened 2023-07-20
Use Meson to detect OpenMP

amontoison:


amontoison:
The culprit are the environment variables in #163.

PR #181, opened 2023-09-29
 Fix an incorrect option in metis5_adapter.c

amontoison:


nimgould (approved):
Very good. I noticed that METIS_DBG_INFO is not set, and currently defaults to 1, which covers the output with metis debug messages. I will change this so that options5[METIS_DBG_INFO] = options[4];
and see if that helps

nimgould:
I get conflicts when I try to merge. No idea why (but then it is git)

amontoison:
> Very good. I noticed that METIS_DBG_INFO is not set, and currently defaults to 1, which covers the output with metis debug messages. I will change this so that options5[METIS_DBG_INFO] = options[4]; and see if that helps

Hi Nick!
`METIS_DBG_INFO` is set [here](https://github.com/ralna/GALAHAD/pull/181/files#diff-0111bf55726ef96f0d5f61a48a7971327920e48a0649b1d0dde1fe557c601b9eL43).
I can fix the git issue.

nimgould:
That statement sets METIS_OPTION_DBGLVL not METIS_DBG_INFO. The C code seems to use METIS_DBG_INFO without reference to METIS_OPTION_DBGLVL, as far as I can see

amontoison:
> That statement sets METIS_OPTION_DBGLVL not METIS_DBG_INFO. The C code seems to use METIS_DBG_INFO without reference to METIS_OPTION_DBGLVL, as far as I can see

Oh you're right! Ping me if you find a way to use the option `METIS_DBG_INFO`. I will also add in libHSL / Coin-HSL.

PR #193, opened 2023-10-11
[CI] Compile GALAHAD with CUTEst

amontoison:
We should `dlopen` the shared library like Ipopt to avoid the issue with missing symbols.


amontoison:
Sorry @nimgould, more bugs to fix for you.
We need just modifications to compile GALAHAD with the 64-bit integer version of CUTEst.

PR #194, opened 2023-10-22
[Meson] Add the clls package

amontoison:


amontoison:
@nimgould The test `clls2` should be updated.

PR #195, opened 2023-11-23
Test the option -DGALAHAD_64BIT_INTEGER

amontoison:
@nimgould  @jfowkes 
I added an option `galahad_int64` in the Meson build system to compile with the flag  `-DGALAHAD_64BIT_INTEGER`.
Should I use a different name of does it make sense to you?

I also added a CI build that compiles BLAS and LAPACK with 64 bits integer such that we can test this option with continuous integration.

Note that we have many errors with this option (see [here](https://github.com/ralna/GALAHAD/suites/18429890758/artifacts/1068313560)).

### DIFF
--- a/.github/workflows/standalone.yml
+++ b/.github/workflows/standalone.yml
@@ -18,12 +18,22 @@ jobs:
             cc_cmd: gcc
             fc_cmd: gfortran
             cxx_cmd: g++
+            int64: false
+          - os: ubuntu-latest
+            compiler: gnu
+            compiler_version: 13
+            cc_cmd: gcc
+            fc_cmd: gfortran
+            cxx_cmd: g++
+            int64: true
+            allow_failure: true
           - os: ubuntu-latest
             compiler: intel-llvm
             compiler_version: 2023.2.0
             cc_cmd: icx
             fc_cmd: ifort
             cxx_cmd: icpx
+            int64: false
             allow_failure: true
           # - os: ubuntu-latest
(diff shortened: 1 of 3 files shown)


nimgould:
Yup, we have used GALAHAD_64BIT_INTEGER in kinds/kinds.F90 to assign the appreopriate fortran and C integer kinds.
I am not suprised that there are errors, in particular galahad calls many other (not lapack/blas) external packages, and I have no idea how these deal with 64bit integers. In addition, we use ssids, and I very much doubt that (at least) the GPU code there is compatible.

amontoison:
What can we do to have a relevant test with continuous integration?
I can compile MUMPS with 64 bits integer but after I don't know if it will be used by default by GALAHAD.

Another remark about the ILP64 interface of BLAS / LAPACK routines, we generally add a `_64` suffix at the end of the name.
I compiled the reference BLAS and LAPACK without this suffix but we should think about it.

I propose to add a `galahad_blas.h` and a `galahad_lapack.h` with the following content:
```
/*
 * Integer specific API
 */
#ifndef BLAS_SUFFIX
#ifdef GALAHAD_64BIT_INTEGER
#define BLAS_SUFFIX(a) a##_64
#else
#define BLAS_SUFFIX(a) a
#endif
#endif

#define GEMV  BLAS_SUFFIX(GEMV)
#define GEMM BLAS_SUFFIX(GEMM)
```

PR #197, opened 2023-11-28
Release 4.1.1

amontoison:
@nimgould I should be able to cross-compile GALAHAD now. Is it fine for you to do a release `4.1.1`?


nimgould:
OK, confirmed. I am just about done with CUTEst 2.2, and once that is done, I'll update both galahad and sifdecode as these too have significant changes. I suspect that the meson builds will fail after this, but it shouldn't be hard to correct this.

nimgould:
Will do. I suspect it will be ok for you, as you are only using the double
precision version. But who knows!

PR #206, opened 2024-01-03
Preprocess blas_interface.F90 and lapack_interface.F90

amontoison:


amontoison:
An `#include "galahad_blas.h"` was missing in `src/norms/norms.F90`.

nimgould:
#include "galahad_blas.h" is not needed, NRM2 is overloaded and imported from the      USE GALAHAD_BLAS_interface, ONLY : NRM2
line. But you are right, we probably need 
  TWO_NORM = NRM2( n, X, 1_ip_ )
(in fortran, one can cast using the explicit function  INT( 1, KIND = ip_ ), or more simply 1_ip_ )

nimgould:
likewise, it should be  CALL SCAL( n, beta, R, 1_ip_ ) without the for the .h file in mop.

I suspect there will be many others like this in other galahad packages that use lapack and blas

amontoison:
> #include "galahad_blas.h" is not needed, NRM2 is overloaded and imported from the USE GALAHAD_BLAS_interface, ONLY : NRM2 line. But you are right, we probably need TWO_NORM = NRM2( n, X, 1_ip_ ) (in fortran, one can cast using the explicit function INT( 1, KIND = ip_ ), or more simply 1_ip_ )

It's needed Nick because you changes the name of the Fortran module for `GALAHAD_BLAS_interface_64`.

nimgould:
I think it is not this, more that I didn't add 
#ifdef GALAHAD_64BIT_INTEGER
#define GALAHAD_BLAS_interface GALAHAD_BLAS_interface_64
#define GALAHAD_LAPACK_interface GALAHAD_LAPACK_interface_64
#endif
to galahad_modules.h. But I could be wrong.

amontoison:
> I think it is not this, more that I didn't add #ifdef GALAHAD_64BIT_INTEGER #define GALAHAD_BLAS_interface GALAHAD_BLAS_interface_64 #define GALAHAD_LAPACK_interface GALAHAD_LAPACK_interface_64 #endif to galahad_modules.h. But I could be wrong.

Yes, but you added it to "galahad_blas.h" and "galahad_lapack.h", whcih are used to compile `blas_interface.F90` and `lapack_interface.F90`.

nimgould:
Indeed, I forgot to add it to galahad_modules at the same time

amontoison:
@nimgould 
Can you help me with the following error?
https://github.com/ralna/GALAHAD/actions/runs/7413624892/job/20172968389?pr=206

nimgould:
I'll have a look first thing tomorrow. The first sort of error is simply that an old-style
abs (iabs) function is called, which presumably didn't make it to 64 bit integer, this
should be easy. The other is because I haven't worked on the dummy hsl routines
to make sure that they accept 64 bit integers yet

nimgould:
I commited an ujpdate, please try again

PR #214, opened 2024-01-08
More documentation about Meson and GALAHAD.jl

amontoison:
@nimgould 
Nick, I've finally taken the time to write documentation on Meson and the Julia interface GALAHAD.jl, as you requested.


nimgould (approved):
Superficially, this is very good. I will have to use it in anger, but in the interim, lI'll commit it

PR #224, opened 2024-01-12
Try to not compile ssids on some compilers

amontoison:


nimgould:
Just to mention that I tried ssids (galahad version) with the nvidia suite, and I can't compile one of the gpu codes (internal compiler error). However, compiling with cuda_nocuda.F90 instead of cuda.F90 (i.e, turn off GPU support) compiles and runs fine. Thus, for the time being, we should do thsi (and the archdefs system has been modified accordingly)

amontoison:
> Just to mention that I tried ssids (galahad version) with the nvidia suite, and I can't compile one of the gpu codes (internal compiler error). However, compiling with cuda_nocuda.F90 instead of cuda.F90 (i.e, turn off GPU support) compiles and runs fine. Thus, for the time being, we should do thsi (and the archdefs system has been modified accordingly)

I opened a PR in the GitHub repository of SPRAL to test it with nvidia compilers.
I have a similar compiler error than `icpc`...
https://github.com/ralna/spral/pull/180

meson.build
```diff
@@ -206,6 +212,20 @@ if libhsl.found()
   add_global_arguments('-DLANCELOT_USE_MA57', language : 'fortran')
 endif
 
+# Link flag for C++
+lstdcpp = '-lstdc++'
+if host_machine.system() != 'linux'
+  if cxx.get_id() == 'clang' or cxx.get_id() == 'clang-cl'
+    lstdcpp = '-lc++'
+  endif
+  if cxx.get_id() == 'intel' or cxx.get_id() == 'intel-cl'
```
amontoison, line 221:
```suggestion
  if cxx.get_id() == 'intel' or cxx.get_id() == 'intel-llvm'
```

amontoison:
I was able to compile GALAHAD (without SSIDS) with the Intel compilers on Mac.
All tests passed :tada: 
I also fixed a bunch of other details.
On Windows, I need to dig more to understand the issue with the Intel linker.

nimgould:
Very good. Do you know what the ssids issue is - what did the compiler grumble about?

amontoison:
Internal compiler error 20000:
https://github.com/ralna/spral/actions/runs/7505963818/job/20436349072#step:10:335
We have this error with all versions of `icpc` (Mac, Linux), and `icl` on Windows.
The new compiler Intel `icx` on Windows that compiles both C and C++ files also has this error.

PR #226, opened 2024-01-15
Improve the Meson build system for 64-bit integer

amontoison:
@nimgould I simplified the Meson build system and added a few comments in the main
`meson.build` such that you, Jari or Dominique can easily understand what represents each variable.

You recently broke the compilation by adding `#include "galahad_hsl.h"` in the dummy HSL files.
Because the extension is `.f90`, it was not preprocessed by default on Linux.
But if I preprocess them, I have an error because the file `galahad_hsl.h` doesn't exist (or you never pushed it on GitHub).
I modified the main `meson.build` such that all `.f90` and `.F90` files are always preprocessed.


nimgould:
oops, sorry, now commited galahad_hsl.h

amontoison:
I propose to change the extension of the dummy files if we preprocess them for 64-bit integer now (`# include "galahad_hsl.h"`).

amontoison:
You probably didn't checked the archive of libHSL that I sent ;)
I don't modify the name of the Fortran modules directly in HSL packages.
When I generate libHSL, I just add an `#libhsl_api.h` with all the macros to add the suffix `_64` to the Fortran modules and BLAS / LAPACK symbols.
The file `libhsl_api.h` is in `libhsl/libhsl` of the branch `libhsl64`.

nimgould:
I had a look, but couldn't understand what you did. The fortran source doesn't have any HSL_INT, so I am unsure how you are prprocessing fortran integer statements. Also C_INT and C_INT64_T are not part of default fortran, you have to use the ISO_C_BINDING module to make those work. Or are you doing something on the C side?

nimgould:
OK, I have now generated the versions that contain the #include "libhsl_api.h"
 at the top (and thus will convert hsl_ma57_double to hsl_ma57_double_64. But
I don't see anything that will convert integer to integer(kind=xx) where xx is the
compiler-dependent 64 bit kind, available from the ISO_FORTRAN_ENV module.

nimgould:
sorry xx = INT64, but only availble from the iso module

amontoison:
In the Meson build system, I added the option:
```
add_global_arguments('-fdefault-integer-8', language : 'fortran')
``
to change the default integer type.
HSL_INT is only for the C interfaces.

nimgould:
That sounds reasonable, but I am rather concerned that a number of the codes use hard-wired 32bit intrinsics. Also some (such as fa04) generate random numbers assuming
32 bit arithmetic, etc, while others (e.g. id05) directly report the base. Thus a global substitution is a very good start, but really each package needs to be checked. if we are distributing this. I am using a similar strategy for the dummy hsl routines (i.e., preprocessing + compiler flags - it is -i8 for nagfor and nvfortran, -qintsize=8 for xlf and -integer-size 64 for ifort/ifx, I haven't checked any others

nimgould:
Damnation. the fdefault-integer-8 flag not only promotes integers to 8 bytes, it does
the same for logicals. Unfortunately, logical kinds are not supported in fortran (at least they do not appear in the iso-fortran module). So unless a user is careful, calls to   MA77_factor_double_64 will result in "There is no specific subroutine for the generic " error message. I have tried LOGICAL( KIND = INT64 ) and this seems to work with gfortran, but who knows if it will with other compilers.

PR #234, opened 2024-01-24
Fix lapack/meson.build

### LINKED ISSUE #233: all builds now fail as blas/lapack are now modern (.f90) rather than traditional (*.f)
The actions all fail as it appears that, as blas/lapack.f -> blas/lapack.f90, the builds still presume that this codes are old source form, and mark the new (free format) continuation statements & as errors. I couldn't find where to change this in the meson build system (maybe it is in actions somewhere instead?)

amontoison:
close #233

@nimgould 
`libgalahad_f_src` is for `*.f` files.
`libgalahad_src` is for `*.f90` or `*.F90` files.
I explained it [here](https://github.com/ralna/GALAHAD/blob/master/meson.build#L168-L214).

Should I also add it in the `README.meson`?

### DIFF
--- a/src/lapack/meson.build
+++ b/src/lapack/meson.build
@@ -2,15 +2,15 @@ libgalahad_src += files('blas_interface.F90', 'lapack_interface.F90')
 
 if not libblas.found()
   warning('building our own BLAS v3.9.1; consider providing an optimized BLAS library')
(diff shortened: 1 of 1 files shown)


nimgould:
Thank you. No I was unaware of this subltety, so maybe a note in the Readme.

PR #235, opened 2024-01-25
Update README.meson

amontoison:


nimgould:
Some of these have overwritten changes that I already modified, and some are identical to changes I already made

nimgould:
Please wait, I don't think it helps if we are both modifying fortran files

amontoison:
@nimgould You pushed after me on `master` the same 
Modfications. I fixed the double.modifications with this PR.

PR #236, opened 2024-01-25
[CI] Add macros.yml

amontoison:
@nimgould 
I added a CI build to check if we have more than 72 / 80 characters on a Fortran line.
It also checks the length of each line when we apply the macros. 

https://github.com/ralna/GALAHAD/actions/runs/7655740194/job/20862507656?pr=236#step:4:5


nimgould:
This sounds good

PR #240, opened 2024-01-29
Add macros for dummy routines of external dependencies

amontoison:
I still need to do the same modifications for the dummy symbols of HSL.


amontoison:
@nimgould,
I removed the macro `GALAHAD_NO_SYMBOL_64BIT_INTEGER` in `galahad_blas.h` and `galahad_lapack.h`.
We encountered an infinite recursion if we defined things like `#define DASUM DASUM`.
If we desire the previous behavior with the suffix `_64` by default, we must always use `-DGALAHAD_UNDERSCORE_64BIT_INTEGER`.
I updated the Meson build system to incorporate this behavior by default.

The current possibilities for the suffix of BLAS and LAPACK symbols are as follows:
- `GALAHAD_NO_UNDERSCORE_64BIT_INTEGER` (suffix `64`)
- `GALAHAD_UNDERSCORE_64BIT_INTEGER` (suffix `_64`)
- `GALAHAD_DOUBLE_UNDERSCORE_64BIT_INTEGER` (suffix `__64`)
- If none of the three macros are used, the BLAS / LAPACK symbols won't have a suffix.

Please note that currently only `-DGALAHAD_UNDERSCORE_64BIT_INTEGER` is operational if we compile with 64-bit integers (flags `-DGALAHAD_64BIT_INTEGER` and `-DSPRAL_64BIT_INTEGER`) because the suffix `_64` is still hardcoded in some files like `src/ssids/cpu_iface.F90` or `src/ssids/cpu_solve.F90`.

amontoison:
Good idea! Let's do that.
I will update the PR tomorrow morning.
Time to go to bed here in Canada.

nimgould:
Great. Thank you. I'll delay the merge until I hear from you. I am working my way through the tests at the moment; fortunately as it is only affects integers, I am not
expecting (nor seeing) any significant behaviour (unless I make a blunder and accidentally delete a line!)

nimgould:
As for HSL, I believe that the decision is to append _64 to modules (but not change the names of prodecures within the modules as those can be changed on a use statement), but we'll need to add a _64 to all non-fortran-90 subroutines. Of course, the latter both breaks the fortran 77 standard (names have to be <= 6 chars), and may push lines above the 72 character limit (so we need to pick compiler flags not to enforce strict 77 and to extend line lengths to accommodate both procedure name changes and 64-bit lapack/blas).

nimgould:
@AndrewLister-STFC @tyronerees may have a view on the above!

AndrewLister-STFC:
> As for HSL, I believe that the decision is to append _64 to modules (but not change the names of prodecures within the modules as those can be changed on a use statement), but we'll need to add a _64 to all non-fortran-90 subroutines. Of course, the latter both breaks the fortran 77 standard (names have to be <= 6 chars), and may push lines above the 72 character limit (so we need to pick compiler flags not to enforce strict 77 and to extend line lengths to accommodate both procedure name changes and 64-bit lapack/blas).

`_64` suffix:
I don't think it matters if the procedures have a suffix -- they're exposed via the confusing use of an interface block so the find and replace won't spot them anyway!

Compiler flags:
As far as I'm aware everything is currently compiled with the same compiler (and flags) whether they are "fortran 77" or not... so I doubt that last point will be an issue for the CI/CD. For users who want to build it themselves, since it's only being updated for libHSL, it would be very hard to compile the "fortran 77" bits with those flags.

amontoison:
@nimgould @jfowkes 
I updated `galahad_modules.h` such that the symbols of the dummy routines and modules of external dependencies have a prefix `galahad_` when we compile them with Meson in one unified shared library.
 
I also reorganized `galahad_modules.h` such that the macros for the dependencies are at the end of the file.

nimgould:
Excellent, thank you

amontoison:
> `_64` suffix: I don't think it matters if the procedures have a suffix -- they're exposed via the confusing use of an interface block so the find and replace won't spot them anyway!

Ok great so this part is already done.
I just need to add the suffix `_64` to F77 and C sysmbols.

> Compiler flags: As far as I'm aware everything is currently compiled with the same compiler (and flags) whether they are "fortran 77" or not... so I doubt that last point will be an issue for the CI/CD. For users who want to build it themselves, since it's only being updated for libHSL, it would be very hard to compile the "fortran 77" bits with those flags.

@AndrewLister-STFC 
The idea is to use this [script](https://github.com/ralna/GALAHAD/blob/master/.github/julia/check_macros.jl) to detect if F77 files have more than 72 characters when we apply the macros.
We will probably need to modify a lot of files, but only once.
Similar to my PR where I renamed the Fortran module to only have lowercase/uppercase names.
And CI can ensure that every time we do a new release of libHSL in the case that you modified an HSL file implemented in F77 (quite rare).

nimgould:
One other possibility is to use the nagfor =polish routine to convert all fortran77 fixed format files to free format. In that way, the 72 char limit rises to 132 chars, and overflows due to preprocessing will "never" happen so long as the converted routines are constrained to (say) 80 chars. The down side is that this will mangle people's codes, and they might not be happy!! I plan to do this in due course, with additional preprocessing commands, for the galahad hsl subset that we ship so that galahad users can compile 64bit hsl from source

PR #243, opened 2024-01-31
Fix more issue for 64-bit integer

amontoison:
- Support 64-bit integer version of METIS;
- Fix a bug with `croti.F90`;
- Don't compile the examples with the 64-bit integer version of GALAHAD.


amontoison:
> I fixed the croti example earlier (same fix) . I don't know how to automatically resolve conflicts, do you?

I rebased the branch :+1: 

The 64bit tests work fine here with the make system (as I diabled ssids) ok for both gfortran and ifx.

We will see it with CI :crossed_fingers:

PR #245, opened 2024-02-04
Add a prefix GALAHAD_ for BLAS and LAPACK symbols when needed

amontoison:
@nimgould 
I tried to use the BLAS interfaces in `ssids/cpu_solve.F90` but it's not working...
I did the same modification in `ssids/cpu_iface.F90` and it works.
Can you have a look on the branch `galahad_blas_lapack`?

https://github.com/ralna/GALAHAD/actions/runs/7775945103/job/21202499421?pr=245#step:12:611

I suspect an issue with the size of the arrays in input of the BLAS routines.


nimgould:
Yes, it must simply be a mismatch for one or more of the arguments. The old cpu_iface.F90 was very specific  about which particular blas routine it uses, while you are now using the interface block to automatically chose - this method needs to find an exact match. All I can suggest is you replace the generic name (DGEMM) with the specific one (sgemm__64 or whatever your procedure is now called), as then the compiler should tell you which argument is causing the problem. I think it is probably because the interface GEMM expects the real array arguments to be 2 dimensional, but lcol and xlocal in cpu_solve are 1D. This is valid fortran as the array is an assumed size one (a * in the length), but not if you are using an interface where the shape is vital. I suggest you go back to the old version of cpu_iface.F90 and put in extra ifdefs to cope  with your renaming of the blas.

nimgould:
I am very amused that meson has spent more than half an hour compiling the code, while the makefile version only took 10 minutes!!

amontoison:
Thanks @nimgould!
It fixed the issue.
I can continue to work on the Julia interfaces.

With the Meson build system, we need 4-6 minutes to compile GALAHAD and run all tests (Fortran, C, Python and Julia).
The build with 64-bit integer takes half an hour because all C tests are not working and many tests are broken and take 30 seconds before that Meson kills them.

nimgould:
Great. We will see. 

On another matter, all of the macos-latest/intel-classic/2021 python tests are failing, so should we skip these - I have no idea how python interacts with fortran/c on this architecture?

PR #246, opened 2024-02-06
Update the name of symbols and Fortran modules

amontoison:
@nimgould
I found a few symbols with `nm -D libhalahad_precision.so` that were not correctly preprocessed.
I fixed them.

Note that a few modules still don't have the `_64` suffix:
- galahad_kinds.mod
- spmf_interfaces.mod
- spral_ssids_blas_iface.mod
- mkl_pardiso.mod
- spral_core_analyse.mod
- spral_ssids_gpu_cpu_solve_double.mod
- mkl_pardiso_private.mod
- spral_hw_topology.mod
- spral_ssids_lapack_iface.mod
- pastixf_enums.mod
- spral_kinds.mod
- spral_ssids_profile.mod
- pastixf_interfaces.mod
- spral_metis_wrapper.mod
- spmf_enums.mod
- spral_pgm.mod


nimgould (approved):
These look fine (superficially). I will only know for sure when I compile using the makemaster system. Thank you for finding the repetitions.

nimgould:
Some of the missing symbols are for packages that mix 32 and 64 bit integers (and are not preprocessed - galahad_kinds and spral kinds are two examples). The others are simply because I don't know what the corresponding external 64 bit modules are called

nimgould:
FYI. The lapack procedures with names larger than 6 chars violate the f77 standard. I didn't know that lapack allowed this, and my rebuild script assumed <= 6 chars. Now fixed!

amontoison:
@nimgould For the module `galahad_kinds`, `ip_` and `ipc_` are different if we use the flag `INTEGER_64`.
Don't we need two distincts modules for 32-bit /64-bit integers?!

https://github.com/ralna/GALAHAD/blob/master/src/kinds/kinds.F90#L44

nimgould:
Yes, sorry, you are right. So the strings SPRAL_KINDS and GALAHAD_KINDS need to be preprocessed in galahad_modules.h. I will do this, and check those other spral modules

PR #251, opened 2024-02-08
Try to use shared libraries on Windows with Intel compilers

amontoison:


amontoison:
Ok, we must use static libraries on Windows with Intel compilers.
I don't know if the issue is with the Intel linker or Meson but let's not spend more time on that.

PR #254, opened 2024-02-09
Update C tests and examples for 64-bit integer

amontoison:


amontoison:
@nimgould I forgot that we also need to update all header files to replace `int` by `ipc_` :skull:

nimgould:
Let's see what happens on the master branch ... thanks for all of this

PR #255, opened 2024-02-09
Update header files for 64-bit integer

amontoison:


include/galahad_arc.h
```diff
@@ -995,7 +995,7 @@ void arc_solve_without_mat( void **data,
  @param eval_f is a user-supplied function that must have the following
    signature:
    \code
-        int eval_f( int n, const double x[], double *f, const void *userdata )
```
amontoison, line 998: @nimgould, I noticed that you use `double` in some function signatures.
Did you forget to replace them by `real_wp_`, or is it intentional?
Should I just use `int` in this context, or is keeping `ipc_` acceptable?

amontoison:
Updating the headers almost fixed all C tests. :muscle: 
We just have two remaining errors (related to `sblstf`) :tada:

include/galahad_arc.h
nimgould: Yes, I think this was only for the docs (I do the same in the docs/C/rst files). This should be changed at some stage. I believe that the codes themselves use real_wp_. All of the docs need to replace int with ipc_ as well., these all predate the 64 bit extension.

nimgould:
I'll give it a go. That sounds like very good news. I've forgotten where I get the output from the sblstf runs on github ... I know you have told me before!

PR #261, opened 2024-02-14
[CI] Add symbols.yml and modules.yml

amontoison:
@jfowkes @nimgould 
I add two new CI builds with this pull request.

The goal of the first new build (`symbols.yml`) is to check if the following shared libraries export the same symbols:
- libgalahad_single.so with 32-bit integer
- libgalahad_double.so with 32-bit integer
- libgalahad_single.so with 64-bit integer
- libgalahad_double.so with 64-bit integer

Two libraries can export the same symbols ONLY IF it's exactly the same routine.
For example `libgalahad_single.so` and `libgalahad_double.so` can both export `iamax_` ONLY IF both libraries are
compiled with 32-bit integer.

Note that I don't compile GALAHAD with SSIDS for now but it seems that many C++ symbols are shared by both single / double libraries.
We probably need to preprocess some symbols in C++ files...

The goal of the second new build (`modules.yml`) is to check that each library generates Fortran modules with different names.


amontoison:
An example of what we can conclude from the output of the two new builds:
- modules.yml : https://github.com/ralna/GALAHAD/actions/runs/7897536131/job/21553279373?pr=261#step:9:20
We need to add macros for HSL modules with 64-bit integer.

- symbols.yml : https://github.com/ralna/GALAHAD/actions/runs/7897536131/job/21553279373?pr=261#step:9:36
`bsc_information` and `fit_information` don't have the suffix `_s` in single precision.

amontoison:
@nimgould Can you confirm that the following Fortran modules are the same in single and double precision?
- "galahad_blas_interface"
- "galahad_lapack_interface"
- "spral_ssids_blas_iface"
- "spral_ssids_lapack_iface"
- "galahad_clock"
- "galahad_string"
- "galahad_symbols"
- "galahad_tools"
- "galahad_copyright"
- "galahad_hash"
- "galahad_common_ciface"
- "galahad_hash_ciface"
- "galahad_hsl_kb22_long_integer"
- "galahad_hsl_mc68_integer"
- "galahad_hsl_mc78_integer"
- "galahad_hsl_of01_integer"
- "galahad_hsl_zb01_integer"
- "galahad_hsl_mc68_integer_ciface"
- "galahad_kinds"
- "galahad_kinds_single"
- "galahad_kinds_double"
- "spral_kinds"
- "spral_kinds_single"
- "spral_kinds_double"
- "mkl_pardiso" (?)
- "mkl_pardiso_private" (?)
- "lancelot_hsl_routines" (?)
- "pastixf_enums" (?)
- "pastixf_interfaces" (?)
- "spmf_enums" (?)
- "spmf_interfaces" (?)
- "spral_core_analyse" (?)
- "spral_hw_topology" (?)
- "spral_metis_wrapper" (?)
- "spral_pgm" (?)
- "spral_ssids_profile" (?)

I'm not sur for the last ones.

amontoison:
@nimgould I checked the modules of my previous message tonight and the modules `pastixf_interfaces` and `spmf_interfaces` are different in single and double precision.

They should have a different name to not collapse if a library / binary is linked with both `-lgalahad_single` and `-lgalahad_double`.

The other modules are only using integers so it's fine if they have the same name in single / double precision.

amontoison:
Nick, we renamed the Fortran modules for 64-bit integer but we did nothing for the C routines.
Should we also add a suffix `64`?

nimgould:
Yes to both of these, I will sort them out as soon as I have a working draft of the HSL subset library.

nimgould:
I am not sure I know what to do with the pastix/spm interface modules as these are provided by the pastix people with the same name for single and double. I can change the galahad dummy names, but what happens when somebody links against a binary library compiled by someone else? Any ideas?

nimgould:
I could presume that the single precision module is  pastixf_interfaces_single, while the double one is just pastixf_interfaces (without a _double)?

PR #301, opened 2024-06-15
Allow positional or keyword arguments in Python interfaces

### LINKED ISSUE #296: Python solve routines should allow keyword arguments
Currently all the Python solve routines accept positional-only arguments (as Python calls it), i.e.
```Python
arc.solve(2, np.array([1.,1.]))
```
As requested by @AndrewLister-STFC, it is more Pythonic to also support keyword arguments when the arguments themselves are non-trivial or when there are a lot of them and it is difficult to keep track, i.e.
```Python
arc.solve(n=2, x=np.array([1.,1.]))
```
is much friendlier to the Python user.

In order to make this change we need to upgrade the parsing in all package solve routines, from:
```C
static PyObject* py_arc_solve(PyObject *self, PyObject *args){
...
// Parse positional arguments
    if(!PyArg_ParseTuple(args, "iiOOOO", &n, &H_ne, &py_x, &temp_f, &temp_g, &temp_h))
        return NULL;
...
}

/* arc python module method table */
static PyMethodDef arc_module_methods[] = {
...
    {"solve", (PyCFunction) py_arc_solve, METH_VARARGS, NULL},
...
};
```
to support positional-or-keyword arguments (as Python officially terms it):
```C
static PyObject* py_arc_solve(PyObject *self, PyObject *args, PyObject *keywds){
...
// Parse positional or keyword arguments
static char *kwlist[] = {"n", "H_ne", "x", "eval_f", "eval_g", "eval_h", NULL};
    if(!PyArg_ParseTupleAndKeywords(args, keywds, "iiOOOO", kwlist, &n, &H_ne, &py_x, &temp_f, &temp_g, &temp_h))
        return NULL;
...
}

/* arc python module method table */
static PyMethodDef arc_module_methods[] = {
...
    {"solve", (PyCFunction) py_arc_solve, METH_VARARGS | METH_KEYWORDS, NULL},
...
};
```

I am happy to implement this at some point when I have some free time (as really I should have templated it this way to begin with), but this will not be anytime soon I'm afraid.

References: [How does PyArg_ParseTupleAndKeywords work?](https://stackoverflow.com/questions/10625865/how-does-pyarg-parsetupleandkeywords-work)

jfowkes:
Resolves #296

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
+          python-version: '3.12'
 
       - name: Install Meson and Ninja
         run: pip install meson ninja numpy
--- a/src/arc/Python/arc_pyiface.c
+++ b/src/arc/Python/arc_pyiface.c
@@ -800,7 +800,7 @@ static PyObject* py_arc_load(PyObject *self, PyObject *args, PyObject *keywds){
 
 //  *-*-*-*-*-*-*-*-*-*-   ARC_SOLVE   -*-*-*-*-*-*-*-*
 
-static PyObject* py_arc_solve(PyObject *self, PyObject *args){
+static PyObject* py_arc_solve(PyObject *self, PyObject *args, PyObject *keywds){
     PyArrayObject *py_x;
     PyObject *temp_f, *temp_g, *temp_h;
     double *x;
@@ -811,8 +811,9 @@ static PyObject* py_arc_solve(PyObject *self, PyObject *args){
         return NULL;
 
     // Parse positional arguments
-    if(!PyArg_ParseTuple(args, "iiOOOO", &n, &H_ne, &py_x,
-                         &temp_f, &temp_g, &temp_h))
+    static char *kwlist[] = {"n","H_ne","x","eval_f","eval_g","eval_h",NULL};
+    if(!PyArg_ParseTupleAndKeywords(args, keywds, "iiOOOO", kwlist, &n, &H_ne, &py_x,
+                                    &temp_f, &temp_g, &temp_h))
         return NULL;
 
     // Check that array inputs are of correct type, size, and shape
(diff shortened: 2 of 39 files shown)


nimgould:
gosh, faster than the hare that Lupin was chasing this morning!

jfowkes:
Ok EQP Python test segfaults, ket me fix that. The rest all pass though which is a good sign.

nimgould (approved):
Thanks, I will merge and update tests for keywords.

Do you have a list of changes to docs?

jfowkes:
That's fixed it, helps if you pass the keyword list.

No sorry, I didn't look through all of them. But the ones listed above all need to be checked so that:
```C
static char *kwlist[] = {"x", "eval_f", NULL};
```
all correspond to what the docs say are the function arguments.

nimgould:
do you have a list of all the keywords, so that I can grep the docs? I'd forgootten that Python variables are case dependent

jfowkes:
If you do a recursive grep for `kwlist[]` you will get all of them.

nimgould:
Have done. Enjoy your weekend, I suspect mine will be adding keywords to tests!

jfowkes:
Thank you! Almost all the tests are now passing!

nimgould:
Yup, it's the Julia one for windows that alexis will inform us about

jfowkes:
Btw it's probably sufficient to modify the existing Python tests to use keyword arguments (unless you also want to test that the argument order is correct when the keywords are omitted). For details see:
https://www.pythondiscord.com/pages/guides/python-guides/parameters-and-arguments/#keyword-arguments

nimgould:
I want to test both as a precaution, just in case someone prefers to use the order option

nimgould:
I now wonder about adding keywords to the test examples as each actually has the arguments with the names of the keywords (I am not saying that it is not a good idea in general where, e.g. n=2 is helpful). A statement such as wcp.load(n=n, m=m, A_type_A_type, A_ne=A_ne, A_row=A_row, A_col=A_col, A_ptr=A_ptr, options=options) seems a bit of overkill. Any feeling?

jfowkes:
Indeed it’s probably not necessary, if the tests currently work they will also work with keywords, all you’d really be testing is if you spelt them correctly in the code.

nimgould:
That is a 0.2 probability event at most. Now working on the docs, and closing this branch

nimgould:
FYI, you missed one, the keyword list for clls_solve_clls. I've added it. 

Python docs now updated (both rst here and html on galahad_docs)

AndrewLister-STFC:
> I now wonder about adding keywords to the test examples as each actually has the arguments with the names of the keywords (I am not saying that it is not a good idea in general where, e.g. n=2 is helpful). A statement such as wcp.load(n=n, m=m, A_type_A_type, A_ne=A_ne, A_row=A_row, A_col=A_col, A_ptr=A_ptr, options=options) seems a bit of overkill. Any feeling?

In python world we generally do this anyway... and then don't worry about the order (although usually names aren't exactly equal).
```
wcp.load(
    n=n,
    m=m,
    A_type=A_type,
    A_ne=A_ne,
    A_row=A_row,
    A_col=A_col,
    A_ptr=A_ptr,
    options=options,
)
```
is the same as
```
wcp.load(
    m=m,
    n=n,
    options=options,
    A_type=A_type,
    A_ne=A_ne,
    A_row=A_row,
    A_col=A_col,
    A_ptr=A_ptr,
)
```

PR #303, opened 2024-06-18
[CirrusCI] Use GCC13 on FreeBSD

amontoison:


amontoison:
We used all our credits for CirrusCI this month. :(

PR #306, opened 2024-06-19
Update hsl_metis.h

amontoison:
I found a message of `svigerske` on an issue of METIS to fix the ABI incompatibilities:
https://github.com/KarypisLab/METIS/issues/71#issuecomment-1696082046


nimgould:
I'll leave this to Jari to test. I only hope that changing the order doesn't have unintended consequences for metis 5.2; I do not know how metis and its header interact.

jfowkes:
> I haven't tested this and prefer to take a flexible approach, which is to allow the user to provide their own version of METIS. When we use `hsl_subset`, we just need to force GALAHAD to compile its own version (`-Dlibmetis=`). I updated the `README.meson` to explain this.
> 
> I also removed an old `metis.h` in the `include` folder so that it is not used instead of the header provided by the user.

Yes this would also be my preference. Note that unlike for MUMPS the reordering of options does not affect `metis5_adapter` as it uses the typedef enums like METIS_OPTION_NUMBERING directly from the #included METIS header, and this is precisely why I wrote the adapter in C and not in Fortran.

.github/workflows/meson.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-13]
         arch: ['x64']
-        version: ['13']
+        version: ['11']
```
jfowkes, line 16: Can we keep GCC 13 please

jfowkes (changes requested):

.github/workflows/meson.yml
amontoison: You need a Julia version compiled with GCC 13 and it's not yet the case on Windows. Is it fine if we use GCC13 except Windows where we keep GCC 11?

jfowkes, line 16:
```suggestion
        version: ['13']
```

jfowkes: Ah okay yeah sure that's fine!

amontoison: I will change in another PR. Csn I merge Jari?

jfowkes: Sure sure

jfowkes (approved):

PR #307, opened 2024-06-19
Test the Julia interface with -Dhsl_legacy=true

amontoison:
`GALAHAD_jll.jl` will be compiled with this option.


amontoison:
@nimgould Something is wrong when we compile GALAHAD with `-Dhsl_legacy=true` on Windows and Mac.

nimgould:
Yes, I need to update ssids_rip.hxx to make it consistent with hsl_precision.h . I will push the change

nimgould:
I have tested and pushed

jfowkes (approved):

PR #308, opened 2024-06-21
Upgrade GHA scripts to be more modular

amontoison:


jfowkes (approved):
Looks good thanks! Still failing for Julia on Windows...

amontoison:
I will duplicate the builds on Windows to see if it's related to the flag `-Dhsl_legacy=true`.
Nick also updated the SPRAL headers to use `Clong` but we should always use `int64_t` there.
`Clong` should be used only in HSL files.

jfowkes:
Yes SPRAL got migrated over to `int64_t` ages ago and the SPRAL headers should definitely not be using `Clong`.

nimgould:
Errrm, when we tested without this fix a couple of days ago, it failed ...

jfowkes:
Maybe you need to update your version of SPRAL to the latest.

amontoison:
@jfowkes @nimgould I duplicated the tests to verify if it's related or not.
We have `Legacy-false` or `Legacy-true` at the end of the build names.

jfowkes:
> galahad spral uses its own variable clong_ which is set to c_int64_t

Ah I see that's fine then (if a little misleading).

nimgould:
The change I made the other day was to the variable longc_ (in ssids_rip.h). I will undo that, but not right now as I am battling with the latest MUMPS!

amontoison:
@jfowkes @nimgould I confirm that the issue is `Clong` on Windows.
All tests passed if we use `-Dhsl_legacy=false` on Windows.

jfowkes:
> Won't that cause all sorts of issues with bindc if long integers are required on the fortran side?

Quite possibly yes.

amontoison:
> > @jfowkes @nimgould I confirm that the issue is `Clong` on Windows. All tests passed if we use `-Dhsl_legacy=false` on Windows.
> > Something is GALAHAD doesn't like the 32-bit integer `longc_ = Clong` on this platform . If we want to reproduce it on Linux we just need to use `longc_ = Cint` when `-Dhsl_legacy=true`.
> 
> Thanks @amontoison, glad we're getting to the bottom of this...
> 
> Won't that cause all sorts of issues with bindc if long integers are required on the fortran side?

You are both more expert than me on Fortran :)

nimgould:
I must say that I am rather reluctant to spend more time on a broken Windows system when evreybody here has moved to WSL. To downgrade every galahad package from long to int will cause all sorts of problems, specifically generic interfaces that rely on the argument lists of
candidates to differ in some aspect (which they won't if long and int are the same). In addition, any link we have to external packages may break, fine for HSL as we are in control, but what about all of the other solvers we support?

I forget, Alexis, are we planning to offer an hsl_subset binary as well as libhsl?

amontoison:
> I must say that I am rather reluctant to spend more time on a broken Windows system when evreybody here has moved to WSL. To downgrade every galahad package from long to int will cause all sorts of problems, specifically generic interfaces that rely on the argument lists of candidates to differ in some aspect (which they won't if long and int are the same). In addition, any link we have to external packages may break, fine for HSL as we are in control, but what about all of the other solvers we support?

Nobody use Python or Julia in the WSL on Windows Nick.
We are doing that for them.

> I forget, Alexis, are we planning to offer an hsl_subset binary as well as libhsl?

That's the plan but I just need to add precompiled library of hsl_subset in HSL_jll.jl.
If we can do that, all problems are fixed.

nimgould:
Since the 64bit version works under Windows, why don't we simply say that this is all
we will support?

I am sorry to be so ignorant, but I am not sure I understand what you mean by adding hsl_subset in HSL_jll.jl. Why isn't this a standalone hsl_subject.jl without the burden of the irrelevant other HSL packages, and with the benefits of being natively 32- and 64-bit capabale?

nimgould:
typedef int64_t longc now restored in ssids_rip.h, and proper mumps tested ok (theer was a subtle bug on my side that needed squashing!)

amontoison:
> Since the 64bit version works under Windows, why don't we simply say that this is all we will support?
> I am sorry to be so ignorant, but I am not sure I understand what you mean by adding hsl_subset in HSL_jll.jl. Why isn't this a standalone hsl_subset.jl without the burden of the irrelevant other HSL packages, and with the benefits of being natively 32- and 64-bit capable?

The precompiled version of GALAHAD in the Julia ecosystem is `GALAHAD_jll.jl`.
It contains `libgalahad_single.$dlext` and `libgalahad_double.$dlext` for all platforms.

If we want to use HSL packages in the Julia interface `GALAHAD.jl`, `GALAHAD_jll.jl` must be linked to a library that contains the HSL symbols needed by GALAHAD. To do that, we want `GALAHAD_jll.jl` to depends on `HSL_jll.jl`.
The issue is that `HSL_jll.jl` only contains `libhsl.$dlext` and not `libhsl_subset.$dlext` / `libhsl_subset_64.$dlext`.

amontoison:
@jfowkes @nimgould 
Bingo
```shell
 ../include/ssids_rip.hxx:38:17: error: conflicting types for 'longc_'; have 'int64_t' {aka 'long long int'}
   38 | typedef int64_t longc_;
      |                 ^~~~~~
In file included from ../include/hsl_ma57.h:47,
                 from ../include/galahad_sls.h:348:
../include/hsl_precision.h:42:14: note: previous declaration of 'longc_' with type 'longc_' {aka 'long int'}
   42 | typedef long longc_;
      |              ^~~~~~
```
We keep `longc_ = long` only for HSL packages so maybe we can rename it `hsl_longc_` to not have any ambiguity.
What do you think?

nimgould:
Aargh ... something you commited has broken everything here; metis is now talking about missing edges.

nimgould:
@amontoison  I believe that you broke things by editing the adapter.c file to give precedence to the system (V 5.1) metis.h.  Really, as this is our package, we should give precedence to our galahad_metis.h. I propose that we change this to SYSTEM_METIS so that by default we take ours, not theirs.

nimgould:
Yep, that was the issue. Can you please redesign your check for metis so that ours is the default.

amontoison:
Ok, so you want that we compile our METIS in GALAHAD by default and allow a custom installation of METIS only as a backdoor?
Note that all CI builds use the GALAHAD version of METIS.

nimgould:
I made all the clong_ -> ssids_clong_ changes, but then undid them (as I thought this caused the issue above!). Do you prefer than we go the hsl_longc_  route instead? I can do either, but I marginally prefer the  hsl_longc_  suggestion

amontoison:
I also prefer `hsl_longc_`.

jfowkes:
Yes we need different names for `longc_` in SSIDS and `longc_` in the HSL headers as otherwise we get the conflicting types issue Alexis mentioned above where one is mapped to `int64_t` and the other to `long` (respectively). I also favour renaming `longc_` in the HSL headers to `hsl_longc_` as it is HSL that uses the now old-school definition of `long`.

jfowkes:
> Ok, so you want that we compile our METIS in GALAHAD by default and allow a custom installation of METIS only as a backdoor? Note that all CI builds use the GALAHAD version of METIS.

Yes I also think this makes sense given that the METIS headers need to be modified to properly support 64bit integers (or at least that is my understanding from looking at the upstream `metis.h` and its hardcoded macros for 64bit integer support).

nimgould:
I've pushed the hsl_longc_ corrections to hsl_subset on the hsl repo, and made similar changes.

I need @montoison to rework src/metis/adapter.c so that the default metis include is galahad_metis.h not metis.h. I am happy to do this, but need to know what cpp variable meson is providing to accomplish this

nimgould:
I have changed adapter.c so that

#ifdef EXTERNAL_METIS
#include "metis.h" /* provided by the user (MeTiS 5.x) */
#else
#include "galahad_metis.h" /* from MeTiS 5.2 */
#endif

as I need to make progress this morning. meson will need to change to reflect this

amontoison:
> I have changed adapter.c so that
> 
> #ifdef EXTERNAL_METIS #include "metis.h" /* provided by the user (MeTiS 5.x) _/ #else #include "galahad_metis.h" /_ from MeTiS 5.2 */ #endif
> 
> as I need to make progress this morning. meson will need to change to reflect this

Done!

nimgould:
Just an update. Two windows runs runs are stubornly refusing to work on two packages, although if this is an issue I supsect that a third is vulnerable. I do suspect a bug since in a loop the control array is corrupted on the second pass, but maybe that is my misundertanding of C loops (can I assume that the status of all variables is unaltered unless I change them? That is, they don't become undefined after the first pass of a loop?)

The main frustration is that it takes 30-40 minutes using actions to get a single piece of data, namely the result of running the C test on the llst package. The same test here (in Linux world) takes 5 seconds max. I appreciate that this because the whole library has to be rebuilt, and ever test performed, but ideally it would be better to have an action that says "compile the bits of galahad that the C version requires to run the llst test on a single target, say windows, gcc, double). John Reid tells me of the early days of computing, where they had two runs max a day, the punched cards were sent by taxi to a computer 30 miles away, run on the machine there, and then the output brought back by taxi).  Githib-actions is just a C21 version of this!! But until I get a Windows machine, that is all I have.

nimgould:
Also, what does legacy false, legacy true mean in the tests?

amontoison:
The build with `legacy=true` means `hsl_longc_ = Clong`.

nimgould:
OK, thanks, so we only seem to fail with this ... and only occasionally. But I suspect that is because there is then a mismatch with the bindc that links fortran to c for hsl routines.

nimgould:
My suspicion is that the issues we see with legacy=true is simply because of what happens on the fortran side. HSL uses selected_integer_kind(18) as a match for clong on the C side. Hsl_subset is different as it uses hsl_longc_ on the C side, but fortran's longc_ which is int64_t by default. The latter enables it to interact with the rest of Galahad. Do I understand that most libhsl interaction by ipopt, etc only uses the C calls? If that is so, libhsl will do the right thing. But Galahad's interaction is via the Fortran calls, and this may be a mismatch. Of course it would be possible to create fortran arrays in Galahad of type selected_integer_kind(18), and copy data, but that would be inefficient. Galahad and hsl_subset are designed to be compatible, while Galahad and libhsl aren't

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

jfowkes:
> My suspicion is that the issues we see with legacy=true is simply because of what happens on the fortran side. HSL uses selected_integer_kind(18) as a match for clong on the C side. Hsl_subset is different as it uses hsl_longc_ on the C side, but fortran's longc_ which is int64_t by default. The latter enables it to interact with the rest of Galahad. Do I understand that most libhsl interaction by ipopt, etc only uses the C calls? If that is so, libhsl will do the right thing. But Galahad's interaction is via the Fortran calls, and this may be a mismatch. Of course it would be possible to create fortran arrays in Galahad of type selected_integer_kind(18), and copy data, but that would be inefficient. Galahad and hsl_subset are designed to be compatible, while Galahad and libhsl aren't.

Yes I believe this is precisely the issue. Your understanding is correct, IPOPT only interacts with HSL via the C interfaces. This is a known HSL-Windows long integer bug/feature that I raised years ago, but unfortunately we cannot fix it in HSL/libHSL by using int64_t instead of C long as this would break precisely those C interfaces that all of our IPOPT users rely on. This is because it would require changing the linear solver headers that previous HSL package versions rely on.

amontoison:
I remember that Ipopt uses the Fortran interface of MA57 and the C interfaces for the other ones.

nimgould:
Ma57 doesn't use long integers...

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

jfowkes:
> Ma57 doesn't use long integers... 

Indeed it does not. Since MA57 is Fortran77 it can be called directly from C which is what IPOPT does (although the wrapper HSL_MA57 now has a C interface with some advanced features useful for optimization this is not yet being used by IPOPT).

amontoison:
Ok, so I propose to drop the support of libHSL in GALAHAD and remove the macro and option for `hsl_legacy`.
We don't have the choise and after all it's better if we only have the maintain `hsl_subset`.

We just have to release a new version of HSL_jll.jl that contains `libhsl.$dlext`, `libhsl_subset.$dlext` and `libhsl_subset_64.$dlext`.

I can't imagine how many codes are broken on Windows due this `Clong` issue....


ps: I will be away from computers this week (26 June -- 1 July).
We plan to go to Bar Harbor in the Maine.

nimgould:
Excellent!! That will make our life so much easier.

Have a good trip, Alexis. I'm away for 10 days from Wednesday, enjoying the rain in the south-west of England. Abnormal service will be resumed when we get back :)

PR #312, opened 2024-07-07
Update hsl_mc68i_ciface.f90

amontoison:


src/external/hsl/hsl_mc68/C/hsl_mc68i_ciface.f90
```diff
@@ -45,7 +45,7 @@ subroutine copy_control_in(ccontrol, fcontrol, f_array_in, f_array_out, &
       type(f_mc68_control), intent(out) :: fcontrol
       logical(lp_), intent(out) :: f_array_in
       logical(lp_), intent(out) :: f_array_out
-      integer, intent(out) :: min_l_workspace
+      integer(ipc_), intent(out) :: min_l_workspace
```
amontoison, line 48:
```suggestion
      integer(ip_), intent(out) :: min_l_workspace
```

nimgould: Yes, this should be an integer(ip_) as it is in the hsl_subset codes.

PR #314, opened 2024-07-09
Test GALAHAD with the dummy libHSL

amontoison:


amontoison:
@nimgould 
I have the following error when I try to compile with the dummy `libHSL`.
Should we rename `galahad_unavailable_option` into `hsl_unavailable_option` in the module [HSL_SYMBOLS](https://github.com/ralna/libHSL/blob/main/hsl_subset/galahad/symbols.f90) to solve the problem?
../src/sls/sls.F90:1268:54:
 1268 |            inform%status = GALAHAD_unavailable_option ; RETURN
      |                                                      1
Error: Name 'galahad_unavailable_option' at (1) is an ambiguous reference to 'galahad_unavailable_option' from module 'galahad_symbols_64'
```

nimgould:
I suppose so. The values will always be the same

amontoison:
I confirm that it solved the issue. :+1:

PR #317, opened 2024-07-09
Test GALAHAD with MKL

### LINKED ISSUE #282: MKL-Pardiso with 64-bit integer
I think that Intel added the suffix `_64` so we probably need to adapt our code here in GALAHAD.

cc @nimgould

amontoison:
close #282

### DIFF
--- a/.github/meson/action.yml
+++ b/.github/meson/action.yml
@@ -145,6 +145,18 @@ runs:
         meson compile -C builddir
         meson install -C builddir
 
+    - name: Install MKL
+      if: inputs.os == 'ubuntu-latest' && inputs.compiler == 'intel'
+      shell: bash
+      run: |
+        cd ..
+        mkdir -p deps
+        VERSION_MKL="2024.2.0"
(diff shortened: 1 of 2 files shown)


amontoison:
@nimgould
The builds on Linux with Intel compilers are now using `MKL` for BLAS / LAPACK.
It also means that we can test the linear solver `MKL_PARDISO` with CI.
I was able to fix the interface of `MKL_PARDISO` when we use 64-bit integer (#282).
However I don't know how to fix #121.
This Fortran [file](https://github.com/jmcdaniel43/Reactive_PB_NN_MD/blob/master/src/mkl_modules/mkl_module_src/mkl_pardiso.f90) on another GitHub repository helped me.

nimgould:
Looks sensible. I will try to find out why the single precision version is failing as soon as I can

nimgould:
This looks simply like an mkl library bug. I get exactly the same behaviour outside of galahad (indeed the galahad example is simply a copy of the one from mkl), namely I get the correct answer for doubles, but junk for singles (floats). I thus suggest we do not support mkl_pardiso in single precision.

nimgould:
UPDATE *** no, it was a consequence of the rather silly convention that all control parameters take default values if the first is one, which overrides the one that says use single not double precision. Doh! Fixed

amontoison:
Thanks Nick! 👍

PR #323, opened 2024-08-15
Add a function to provide the version of GALAHAD in the Julia interface

amontoison:
I modified a little bit the Fortran implementation of `VERSION`.


nimgould:
OK, having this as a subroutine rather than a function is not so usual, but I can llive with this.
But I don't want this to be different from other galahad C interfaces, we should have this 
as a standalone C interface module, galahad_version_ciface . Also, really both C and fortran routines should be version_galahad (or something similar) not version (on the fortran side, that is my mistake); version_number would be better, but I suppose there may be other version_numbers in C. Finally comments should say subroutine not function, and the date at the top needs to be current. I am happy to do this once you have merged.

amontoison:
Feel free to do the modifications on my branch Nick:
```shell
git fetch origin
git checkout version_galahad
```
I will stop for tonight.

nimgould:
OK, will do, but only after I've been for a hike, the sun is out

nimgould:
I made the changes, could you check, please

src/version/C/version_ciface.F90
```diff
@@ -0,0 +1,41 @@
...
+!  Copyright reserved, Gould/Orban/Toint, for GALAHAD productions
+!  Principal authors: Jaroslav Fowkes & Nick Gould
+
+!  History -
+!    originally released GALAHAD Version 5.1. August 16th 2024
+
+!  For full documentation, see
+!   http://galahad.rl.ac.uk/galahad-www/specs.html
+
+  MODULE GALAHAD_VERSION_ciface
```
amontoison, line 17: Do we have a macro for `GALAHAD_VERSION_ciface_64`?

src/version/C/version_ciface.F90
```diff
@@ -0,0 +1,41 @@
...
+
+    USE GALAHAD_VERSION, ONLY: f_version_galahad => version_galahad
+
+  CONTAINS
+
+!  --------------------------------------
+!  C interface to fortran version_galahad
+!  --------------------------------------
+
+    SUBROUTINE version_galahad( major_version, minor_version, patch_version )  &
```
amontoison, line 27: Should we add a suffix `_s` like the other routines in single precision?
-> `version_galahad_s`

src/version/version.F90
```diff
@@ -16,33 +16,32 @@
...
 
-      USE GALAHAD_KINDS_precision
-
       IMPLICIT NONE
 
       PRIVATE
-      PUBLIC :: VERSION
+      PUBLIC :: VERSION_galahad
 
       INTEGER, PARAMETER :: major = 5
```
amontoison, line 24: It should be 64-bit integers if we compile GALAHAD with` -DINTEGER_64`.

amontoison (commented):
We probably also need to update `galahad_version.h` to be consistent with the other header files.

src/version/version.F90
nimgould: No, these are tiny integers, no need for 64 bit, and there are no real values so do we need _s?

amontoison: Yes these are tiny integers but we could want 64-bit integers for compatibility reasons with other languages like MATLAB.

For the `_s`, you are right, we don't need it.

nimgould: > Yes these are tiny integers but we could want 64-bit integers for compatibility reasons with other languages like MATLAB.
> 
OK, we can do that. On my list (the sun is out, and I am on vacation, so I'll probably
head out to the garden now)

amontoison:
All tests passed, thank @nimgould !

PR #324, opened 2024-08-16
Update galahad.h

### LINKED ISSUE #196: Update galahad.h?
I am wondering if we should not add a list of `#include` for all packages in the main `galahad.h`.
It's what is done in Intel MKL for example. The user can just add an `#include "mkl.h"` in its code .

amontoison:
close #196


nimgould (approved):
Looks good to me. Thanks

PR #332, opened 2024-11-01
Update the Julia documentation for the future release 0.3.0 of GALAHAD.jl

amontoison:


amontoison:
Nick, the version of the Julia interface is not 1.0. 
You can automatically found it [here](https://github.com/ralna/GALAHAD/blob/master/GALAHAD.jl/Project.toml#L4).

nimgould:
OK, I will process early tomorrow ... its Friday night here, and that only means one thing ...

nimgould (commented):
Looks good to me

nimgould:
New Julia docs have been deployed (with 0.3.0 version number ... do all Julia packages start at version 0?)

amontoison:
No, you can start with 1.0 if you want, but if you versioned your Julia package correctly, you need to bump the major number if you break the API.  
It makes more sense to have a version 0.3 after 0.2 with the latest modifications than to go directly to 2.0 before we decide that it is an official 1.0 release.

We can tag a 1.0 release when the Julia package becomes more stable and won't change too much.  
We don't support Int64 yet in the Julia interface because I need to recompile GALAHAD, and I'm waiting for a new release of the Fortran code to do that.
Maybe we will change the API again fo that.

nimgould:
OK, thanks, that makes sense

PR #335, opened 2024-11-06
[CI] Try GCC 13 for Windows

amontoison:


.github/workflows/meson.yml
```diff
@@ -7,35 +7,13 @@ on:
...
-    name: ${{ matrix.os }}/${{ matrix.compiler }}-v${{ matrix.version }}/Int${{ matrix.int }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: ['ubuntu-latest', 'macos-13']
+        os: ['windows-latest', 'ubuntu-latest', 'macos-13']
         arch: ['x64']
         compiler: ['gcc']
         version: ['13']
```
amontoison, line 19:
```suggestion
        version: ['12']
```

amontoison:
@nimgould It seems that `lancelot_steeringt` in not working on Windows with double precision with GCC 12.
It could be a compiler error but a quick memory could still be relevant.

nimgould:
I tried here in Linuxland with gcc12 and valgrind, nothing suspicious and it runs to completion. So I suspect that it is a Windows issue

PR #336, opened 2024-11-09
Remove old BLAS and LAPACK files

amontoison:
They are not needed anymore by the Meson build system.


nimgould:
I've made a local copy as these are needed if I ever need to regenerate the official blas and lapack subset

PR #339, opened 2024-11-21
Update the headers of hsl_subset

amontoison:
- Add macros for quadruple precision in `galahad_cfunctions.h`.
- Update the headers of `hsl_subset` that were updated recently (only the C routines).


nimgould:
Ermm ... I suspect that this will break everything that I have been working on in the quad branch for the past two weeks! I made changes to all of these files ... could you hold off modifying GALAHAD until I am satisfied that it works at least under make, please

PR #349, opened 2024-12-13
Support quadruple precision in GALAHAD.jl

amontoison:
@jfowkes @nimgould  
This PR adds support for quadruple precision in the Julia interface `GALAHAD.jl`.  
In the same spirit as the unified support for single and double precision, a new precision entry `Float128` is now available and will dispatch to all routines of GALAHAD in quadruple precision.


amontoison:
I have a segmentation fault on Linux:
```julia
in expression starting at /home/runner/work/GALAHAD/GALAHAD/GALAHAD.jl/test/test_glrt.jl:97
glrt_solve at /home/runner/work/GALAHAD/GALAHAD/builddir/../src/glrt/glrt.F90:1521
glrt_solve_problem at /home/runner/work/GALAHAD/GALAHAD/builddir/../src/glrt/glrt.F90:2785
glrt_solve_problem_q at /home/runner/work/GALAHAD/GALAHAD/builddir/../src/glrt/C/glrt_ciface.F90:413
glrt_solve_problem at /home/runner/work/GALAHAD/GALAHAD/GALAHAD.jl/src/wrappers/glrt.jl:118
test_glrt at /home/runner/work/GALAHAD/GALAHAD/GALAHAD.jl/test/test_glrt.jl:57
```

nimgould:
I have no idea without putting write statements into the fortran (and if I do, where would the fortran output be written?) The offending fortran line is a simple vector update        R = R + data%alpha * ( VECTOR + sigma * data%W( : n ) )
so we would need to check all the values R, data%alpha, Vector, sigma and data%W to see which is the source of the error

amontoison:
The Fortran ouptut will be written to stdout so we will see it in the Julia tests / Windows tests.
I'm almost sure that the remaining errors are related to memory.

amontoison:
> I have no idea without putting write statements into the fortran (and if I do, where would the fortran output be written?) The offending fortran line is a simple vector update R = R + data%alpha * ( VECTOR + sigma * data%W( : n ) ) so we would need to check all the values R, data%alpha, Vector, sigma and data%W to see which is the source of the error

Nick, I suspect a bad initialization of `data%W`. Can you check that it's correctly allocated?  
It's the only component of the update where we might access out-of-bound memory.

nimgould:
If we are in this section of the fortran code, data%W should have been allocated to be of length n. But line 1521 can only be accessed if either weight = 0 or power = 2, neither of which is possible for the julia test example you gave. Thus it looks as if something is
not arriving at the fortran end , and is being corrupted between julia and fortran. In the julia call, power = 3.0; does the Julia call know that is supposed to be a real128?

nimgould:
I also ran this through valgrind with no memeory issues reported.

A question: is it possible with actions to run one test with the libraries
needing only the packages involved? This is trivial with the makefile setup, so I presume it is also true for meson? For example, suppose I just want to run the double-precision fortran test of gltr on Ubuntu linux. Rebuiling and retesting everything is clearly wasteful for a single test

amontoison:
> If we are in this section of the fortran code, data%W should have been allocated to be of length n. But line 1521 can only be accessed if either weight = 0 or power = 2, neither of which is possible for the julia test example you gave. Thus it looks as if something is not arriving at the fortran end , and is being corrupted between julia and fortran. In the julia call, power = 3.0; does the Julia call know that is supposed to be a real128?

Yes, but it was not correctly reinterpreted from `Float64` automatically! Passed directly power in `Float128` fixed the issue.
I have a `printf` error now in Julia and I'm wondering if it's normal that we have a so big value for `inform[].obj_regularized`:
```
1.18973149535723176508575932662800702e+4932
```
I have a similar error in the Julia tests of `gltr` and `lstr`.

amontoison:
> I also ran this through valgrind with no memeory issues reported.
> 
> A question: is it possible with actions to run one test with the libraries needing only the packages involved? This is trivial with the makefile setup, so I presume it is also true for meson? For example, suppose I just want to run the double-precision fortran test of gltr on Ubuntu linux. Rebuiling and retesting everything is clearly wasteful for a single test

Meson detects which files were modified since the last build and recompiles only those if we work locally.
However, we need a complete initial build with all the packages of GALAHAD.

With CI, it's harder to fulfill your request because it would require adding some options to the build system to compile only specific components of GALAHAD (first build).
This feature is more suited for developers, and since we can already handle this locally through the build system, I'm not particularly motivated to implement it, as it would take a few hours of work.

nimgould:
OK, I just thought I'd ask. As the main developer, I'm happy just to use make, and leave meson as the deployment system.

The printf error is very similar to those I had in C until I changed from trying to print
128 bit numbers natively (Qf isn't implemented in general). The simplest fix there is to printf a (double) value, and maybe you have to do something  similar in Julia?

amontoison:
I removed the printf just to test if all tests passed and I have now this segmentation fault:
```shell
[204082] signal 11 (1): Erreur de segmentation
in expression starting at /home/alexis/Bureau/git/GALAHAD/GALAHAD.jl/test/test_glrt.jl:101
glrt_trts at /home/alexis/Bureau/git/GALAHAD/builddir/../src/glrt/glrt.F90:2240
glrt_solve at /home/alexis/Bureau/git/GALAHAD/builddir/../src/glrt/glrt.F90:1614
glrt_solve_problem at /home/alexis/Bureau/git/GALAHAD/builddir/../src/glrt/glrt.F90:2785
glrt_solve_problem_q at /home/alexis/Bureau/git/GALAHAD/builddir/../src/glrt/C/glrt_ciface.F90:413
glrt_solve_problem at /home/alexis/Bureau/git/GALAHAD/GALAHAD.jl/src/wrappers/glrt.jl:118
test_glrt at /home/alexis/Bureau/git/GALAHAD/GALAHAD.jl/test/test_glrt.jl:57
```

nimgould:
I have no idea here.  The C and Julia tests are the same, and the C test works fine with fully debugging on. The offending line is the copy OFFD_fact = OFFD, and of course this should be a no-op if n =1 (and fortran allows for this).  Let me make the skip explicit (i.e., IF ( n > 1 ) OFFD_fact = OFFD); I'll commit this and see if it helps you

amontoison:
It doesn't help on Linux but it works on Mac.

It means that we have an internal compiler error or some memory is filled with chuncked and depending on the OS, it could be valid or not.

I updated the CI script to test the compilation with more version of `gcc` / `gfortran`. It will help to isolate the issue.

nimgould:
OK, without putting write statements into the fortran, I doubt we will find it. But when I look at the output (maybe I am looking at the wrong place), I see mention on the Julia side of bad entry 4397 of an array ... for a problem whose arrays should be at most 100. It is also odd that it is all the 32bit integer runs that are failing, and not the 64bit ones

amontoison:
It's because I didn't interfaced the 64-bit interface API in the Julia interface yet.
We still have 24 errors in Windows for the C tests and I don't have any idea what is the issue.
Same errors on various versions of gcc / gfortran, so it's not compiler errors...

Off-topic: @nimgould @jfowkes I can use a self-hosted cluster at Argonne, `moonshot`, to test GALAHAD with non-open source libraries like libHSL or GPU, which could be also useful for the GitHub repository of SPRAL.

nimgould:
The other thing that is odd (to my eye) is that the output talks about 64bit integer prints for a 32bit run.

I'm sorry but without knowing where the Windows errors are occurring, it is impossible to debug

amontoison:
Can you share the link of where you see the output that talks about 64bit integer prints for a 32bit run?

amontoison:
Nick, I may have found something. In `GLRT`, I have a very small value for `inform[].obj_regularized`, which causes the call to `printf` to fail. This value is filled by Julia when it creates the structure that we pass to GALAHAD:  
```  
julia> b = Ref{Float128}()  
Base.RefValue{Float128}(1.49491971601022727328352266985243406e-4932)  
```  
I suspect that `inform[].obj_regularized` is not being updated by the C interface for some reason.

I think the errors on Windows are related as well, because each OS can fill the arrays with whatever chunks they want.
We encountered a similar issue in CUTEst, if you remember:
https://github.com/ralna/CUTEst/issues/63#issuecomment-2310383341

nimgould:
GLRT: Error During Test at /Users/runner/work/GALAHAD/GALAHAD/GALAHAD.jl/test/test_glrt.jl:100
31994
  Test threw exception
31995
  Expression: test_glrt(Float128) == 0
31996
  BoundsError: attempt to access 4937-element Vector{UInt8} at index [4986]
31997
  Stacktrace:
31998
   [1] throw_boundserror(A::Vector{UInt8}, I::Tuple{Int64})
31999
     @ Base ./essentials.jl:14
32000
   [2] setindex!
32001
     @ ./array.jl:986 [inlined]
32002
   [3] format(::Vector{UInt8}, ::Int64, ::Printf.Format{Base.CodeUnits{UInt8, String}, Tuple{Printf.Spec{Val{'i'}}, Printf.Spec{Val{'i'}}, Printf.Spec{Val{'i'}}, Printf.Spec{Val{'f'}}}}, ::Int64, ::Vararg{Any})

amontoison:
This error is because, the value in the structure / reference is not updated and filled with something invalid.
Julia doesn't know how to display it.
I removed the `printf` that we can see where is the error later with a segmentation fault.

nimgould:
Sorry, offline until tomorrow, we have guests staying

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

amontoison:
Don't worry, I have my flight for Paris in a few hours.
Should be back online on Tuesday.

nimgould:
OK, after a couple of long days of debugging, here are my conclusions

1. For glrt, the issue seems to be the julia interface in the real128 case. Namely, the values of power and weight that appear in test_glrt as 3 and 1 arrive on the C side as 0 and 0 (and this then leads to an allocation error if the solver is recalled, as it is in the tests) This does not happen for the C tests, so all I can deduce is that Julia and C are not cooperating for real128s
2. For eall the C interface failures under Windows, the errors seem to occur when we call  quadmath_snprintf for the quadmath library; commenting it out, and the runs succeed. This suggests that this has not been implemented properly under Windows. I can code around each of these by truncating to double, but it does mean that Windows users will not be able to see full real128 answers. Are you sure that you are linking this properly?

PR #356, opened 2024-12-22
[CI] Add a workflow for the code coverage

amontoison:


amontoison:
@jfowkes @nimgould  
I found how to do code coverage with Meson for Fortran code and upload the results to Codecov:  
https://app.codecov.io/gh/ralna/GALAHAD  

I added a badge to the `README` so that we can easily track its evolution.  
Nick, the new workflow uses all our Fortran/C tests to determine which lines in the source code are covered or not.  
Lines not covered are at greater risk of bugs (functions not tested or branches of the code that we never execute).  

@AndrewLister-STFC  
It might interest you to know that we can achieve this with Meson and `gcovr` for potentially the future "SparseHSL".

PR #360, opened 2025-01-02
[Meson] Add an option to compile or not the binaries

amontoison:
@jfowkes @nimgould 
I added an option to the Meson build so that the binaries are no longer compiled by default.
I believe the majority of users are not interested in these files, and making their compilation optional seems appropriate from my point of view.

I will wait a green light of you to merge the PR.
Capture d’écran du 2025-01-03 00-02-50

### DIFF
--- a/.github/meson/action.yml
+++ b/.github/meson/action.yml
@@ -247,6 +247,7 @@ runs:
                              -Dcpp_std=$CPPSTD \
                              -Dexamples=true \
                              -Dtests=true \
+                             -Dbinaries=true \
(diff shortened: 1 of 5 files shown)


nimgould:
I don't know what you mean here, does Julia not need binaries? Sorry for my ignorance

amontoison:
Julia only need the shared libraries, just like Python.
They don't need any executable.

nimgould:
OK, I guess it is just a matter of terminology, but here in fortan-land executables are components of libraries. Again, not needed here except for the main programs

PR #362, opened 2025-01-02
Add the suffix _type for the C structures in GLS

amontoison:
@nimgould Is it fine for you?
It doesn't concern the code in Fortran, it's just the structures in the C interface that are renamed.
I always need some sed commands in the Julia interface because of this missing suffix.
Only `GLS` is concerned by this "issue".


nimgould:
I am not completely happy with this as it is breaking change for any C user who already uses the gls package. Of course, this may be nobody! OK, let's do it, and I'll divert any grumbles to you :)

PR #377, opened 2025-02-08
[GALAHAD.jl] Test the Julia interface with GALAHAD_jll.jl in PRs

amontoison:


amontoison:
@jfowkes @nimgould 
I confirm that we don't have any issue on Mac with SSIDS + `Int32`.
I still don't what could be the issue with `Int64`.

PR #386, opened 2025-04-02
[CI] Use sanitizers

amontoison:


amontoison:
@jfowkes @nimgould 
`ASAN` and `UBSAN` found a few things to fix!

jfowkes:
Thanks @amontoison looks like there are quite a few Fortran memory leaks in some of the GALAHAD codes.

nimgould:
Perhaps you can see something I can't, but where are the memory leaks? Very happy to fix anything, of course.

amontoison:
@nimgould 
You this page: https://github.com/ralna/GALAHAD/actions/runs/14229985592
You can download two archives with the logs of the tests performed by Meson with the sanitizers `ASAN` and `UBSAN`.
For each failing test you have the reason why the sanitizers are not happy.

nimgould:
Thanks @amontoison . I very much doubt I'll be able to fix many of these as they don't fail with tools I have here, and the reports given are too vague to track memory issues. Do you have better insights?

nimgould:
Interestingly, all the errors (bar that for sls) occur with the C interfaces that use external function evaluations. @jfowkes could 
there be an issue here?

nimgould:
I've sorted out a few of these (there were issues!). However, the majority seem to be that the sanitizer is unable to deal with fortran optional keyword arguments. I get no errors with valgrind (or without any mem checks). I propose that we ignore these unless you have any better ideas

amontoison:
@nimgould I checked online and `UBSAN` can return false positive with optional arguments in Fortran.
I remove it in the tests with sanitizers to keep only `ASAN`.

If you want to run it locally, you can use the following commands:
```shell
meson setup builddir --buildtype=debug -Dtests=true -Db_sanitize=address
meson compile -C builddir
meson test -C builddir
```
If you have an error, Meson gives you the full command used in `testlog.txt` (file that you can find `builddir`) and enter the command in your shell directly for more details.

I opened an issue to not forget to document the various sanitizers (#388).
The option for `UBSAN` is `-Db_sanitize=undefined` and the option for `MSAN` is `-Db_sanitize=memory`.
We don't use `MSAN` for now in CI because we need to compile with `clang`, `clang++` and `flang` compilers on linux.

nimgould:
Thanks, Alexis. I'll let you know if I find anything else. As I said, I see no memory issues with valgrind on these examples, and the message from ubsan seems to suggest that it is making a jump in the code as it gets confused by the optional arguments. Of course, I am not ruling out an issue.

jfowkes:
I recall seeing this UBSan behaviour in Fortran before, I think these are genuine false positives, the way we handle optional arguments in Fortran is according to the 2018 standard and this it seems is too new for UBSan to properly support.

amontoison:
I removed `UBSAN`, the remaining errors are only with `ASAN` now.

nimgould:
The ASAN ones seem to be pretty similar, it gets confused by optional arguments on the C side (where a NULL should so the trick according to the fortran standard)

jfowkes:
Again this is a recent addition in the Fortran 2018 standard so I'm not surprised by this. Fortran support for the sanitisers tends to lag behind, this isn't the first time we've found it gets confused by the cutting edge features.

amontoison:
I will not compile the C interfaces in that case for the build with `ASAN`.

PR #398, opened 2025-04-14
Set 1-based indexing inside the routines *_initialize

amontoison:


nimgould:
Well done, that is helpful. I was just waiting for the moonshots to complete before I merged, but they were queued until past my downtime

PR #411, opened 2025-05-02
[TEMP] Build windows wheels

jfowkes:
Temporary CI script to try and build windows wheels for PyPI.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -1,0 +1,84 @@
+name: Build and upload to PyPI
+
+# Build on every branch push, tag push, and pull request change:
+on: [pull_request]
+# Alternatively, to publish when a (published) GitHub Release is created, use the following:
+#on:
+#  release:
+#    types:
+#      - published
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # Note: builds on these are backward compatible
+        # macos-13 is an intel runner, macos-14 is apple silicon
+        os: [windows-latest]
+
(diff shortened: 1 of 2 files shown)


jfowkes:
We'll we're still seeing:
```
NotImplementedError: Bundling libraries in wheel is not supported on win32
```
Relevant discussion about this is https://github.com/mesonbuild/meson-python/discussions/629

However, there doesn't seem to be a solution yet unfortunately.

jfowkes:
Okay we are getting there, from [the relevant meson-python docs](https://mesonbuild.com/meson-python/how-to-guides/shared-libraries.html#internal-shared-libraries) I just need to figure out how to do this:

> If an internal shared library is not only used as part of a Python package, but for example also as a regular shared library then the method shown above won’t work. The library is then marked for installation into the system default `libdir` location. Actually installing into `libdir` isn’t possible with wheels, hence `meson-python` will instead do the following:
>
> 1. Install the shared library to the `.<project-name>.mesonpy.libs` top-level directory in the wheel, which on install will end up in `site-packages`.
> 2. On platforms other than Windows, rewrite RPATH entries for install targets that depend on the shared library to point to that new install location instead.
>
> On platforms other than Windows, this will make the shared library work automatically, with no other action needed from the package author. **On Windows, due to the lack of RPATH support, the `.<project-name>.mesonpy.libs` location search path needs to be added to the DLL search path**, with code similar to the one presented above (this function is very similar to the snippet inserted into the main `__init__.py` of a package by `delvewheel` when it vendors external shared libraries).

It would be nice if `delvewheel` could do this for us, so I will ask if it is something they are willing to add support for.

amontoison:
Wow GG @jfowkes !!! :open_mouth:

PR #414, opened 2025-05-08
Support the binaries for QPLIB in GALAHAD.jl

amontoison:
@nimgould How does QPLIB support work in single and quadruple precision? 
Does GALAHAD read the data in double precision and then perform the conversion?

Can you also confirm that the correct way to call a binary for QPLIB is:
```shell
runbpq_qplib_double < myfile.qplib
````


nimgould:
The qplib file is only available in one precision. Each real is read as * , which is fortran for "convert to the precision of the variable that it is being written to".

Yes, the file is read on unit 5 (standard input) so < from the binary

PR #415, opened 2025-05-08
Automatically build Python wheels on Windows

jfowkes:
Finally with some nasty DLL hacking on Windows we can automatically precompile Python binaries on Windows.


nimgould:
Excellent. Well done!

PR #427, opened 2025-07-11
[Meson] Compile EXPO and SLLS for prototyping

amontoison:


nimgould:
Thanks, Alexis. As you will have discovered, this is very much a work in progress (and relies on a new linear solver SSLS, which is not yet documented). There is a huge amount still to do, so I would not suggest we make this widely available. But, by all means use it yourself. The CUTEst interface works fine, but of course there is no C, Python, Julia, Matlab, docs or anything else useful at the moment.

I'm parallel processing for the time being as I hope to release the new trust-region and norm regularization solvers trek & nrek as well. Again, these need a lot more work, but at least trek is much more polished than expo, and just as exciting.

PR #439, opened 2025-07-20
Test GALAHAD with MA97 on moonshot

amontoison:
I suspect a bug in the interface for MA97, let's try to isolate it.


nimgould:
I'll have a look at this once I have sorted out the ssids issues :)

nimgould:
OK, I have tested all of the above locally, and I get no failures with ma77 or ma97. The "ma77   fail in analyse, status = -43" suggests that direct-access files are not supported on moonshot (that is how ma77 works for large problems, data is written in and out to files). All of qp routines that you report failures (in ma64_swap or factor_solve_block_small) work fine here with gfortran, nagfor and ifort. The line your report a failure ../src/hsl_ma97/hsl_ma97r.f90:8088 is a call to the blas routine *swap, and this suggests that something is up with your blas implementation

amontoison:
@nimgould Moonshot compiles a static version of Reference-LAPACK now (for Int32 or Int64) for `hsl_subset`.
I restarted all builds, let's see if everything works with your recent fixes.

nimgould:
Excellent. So it was an open-blas issue, what a pity since we rely on the implementers of these get things right.

nimgould:
I agree that we should fix those symbol clashes, but all my attempts so far show my limitations with C++. I'll continue to try in the background.  I have been keeping this to the branch ssids_64, if you want to look at the compiler messages (which I don't understand!)

amontoison:
> I agree that we should fix those symbol clashes, but all my attempts so far show my limitations with C++. I'll continue to try in the background. I have been keeping this to the branch ssids_64, if you want to look at the compiler messages (which I don't understand!)

I will have a look asap, probably next Monday after ICCOPT.

PR #448, opened 2025-07-28
Fix CI builds broken by Nick

amontoison:
Run CI during my flight Los Angeles -> Chicago.
I can only commit offline for the next 5 hours.


nimgould:
Thanks, Alexis. I wondered why the CI couldn't find those extra ssls and expo types. Now I see I needed to add them in the rewriter and GALAHAD julia files too!

I also noticed that one on expo's control parameters (stop_advance_start should be stop_advanced_start) has a typo (everywhere). I will fix once I pull at this end

nimgould:
Just a minor comment. I try to keep all text/codes in galahad files to fit in 80 characters (this is the default window width in some applications). The Julia files mostly overflow this. Is this necessary? Or can it be fixed automatically?

nimgould:
Oh, your fixes didn't all work. You removed the T from the ssls_control_type, but this has a component sls_control_type which requires T. Could you fix this one, please (I don't know if this was generated by a macro, or by hand)

nimgould:
Maybe too late, I merged it, sorry, I thought you had put the ball back into my court

amontoison:
I just opened #449, it contains what I did in the plane.

I know that the Fortran tests of SSLS don't compile and I also have errors for EXPO. I will open another PR to not include new failures on `master`.

amontoison:
> Just a minor comment. I try to keep all text/codes in galahad files to fit in 80 characters (this is the default window width in some applications). The Julia files mostly overflow this. Is this necessary? Or can it be fixed automatically?

It is not necessary, it is just that I don't use [JuliaFormatter.jl](https://github.com/domluna/JuliaFormatter.jl) automatically.
We can easily fix that. I only use the formatter when I generate the Julia wrappers from the C interface.

nimgould:
That would be good, it would make the codes look more "professional", and easier to read/maintain

PR #449, opened 2025-07-28
Fixes for EXPO and SSLS

amontoison:


nimgould:
I'll fix the usessls errors

amontoison:
For EXPO, what would you like that we test / don't test with CI?

nimgould:
I think it would be good to do this as a sanity test, but as I said, I anticipate this changing a lot over the next year as we refine the algorithm

PR #459, opened 2025-07-30
[CI] Build python wheels on version tags

### LINKED ISSUE #457: build_wheels.yml is not triggered by new releases
We still have the version `5.2.0` on PyPI.
@jfowkes Do you use the precompiled binaries that we generate with BinaryBuilder.jl for PyPI or do you have your own compiler / build system for PyPI?
I think the issue is related to this: https://github.com/ralna/GALAHAD/blob/master/.github/workflows/build_wheels.yml#L9
We should maybe start it on new "tags".

jfowkes:
Resolves #457

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -1,12 +1,10 @@
 name: Build and upload to PyPI
 
-# Build on every branch push, tag push, and pull request change:
-#on: [pull_request]
-# Alternatively, to publish when a (published) GitHub Release is created, use the following:
+# Publish when a new GitHub version is tagged:
 on:
-  release:
-    types:
-      - published
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -1,12 +1,10 @@
...
-#on: [pull_request]
-# Alternatively, to publish when a (published) GitHub Release is created, use the following:
+# Publish when a new GitHub version is tagged:
 on:
-  release:
-    types:
-      - published
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
```
amontoison, line 7: I trust you for the regex :)
I use `'v*'` for the other workflows but it should give the same result.

jfowkes: Haha I took it form here:
https://stackoverflow.com/a/71076957
Feel free to change to `'v*'` if you prefer.

Issue #458, opened 2025-07-30
[documentation] Explain how to use precompiled binaries


amontoison:
I can recycle what I did in the documentation of `libHSL` or `Uno`:
https://github.com/cvanaret/Uno/blob/main/INSTALL.md#precompiled-libraries-and-executables

PR #462, opened 2025-07-31
[documentation] Explain how to use precompiled binaries and executables

amontoison:
close #458


nimgould (approved):
Again, seems reasonable

PR #463, opened 2025-07-31
[documentation] Explain how to switch BLAS / LAPACK with LBT in GALAHAD.jl

### LINKED ISSUE #394: [documentation] Explain how to switch BLAS / LAPACK with LBT in GALAHAD.Jl
I wrote a section about it in the README.md of `Ipopt.jl`: https://github.com/jump-dev/Ipopt.jl?tab=readme-ov-file#blas-and-lapack
It should also be in the README.md of `GALAHAD.jl` and `HSL.jl`.

amontoison:
close #394

### DIFF
--- a/GALAHAD.jl/README.md
+++ b/GALAHAD.jl/README.md
@@ -13,6 +13,10 @@ pkg> test GALAHAD
 If you launch Julia from within the folder `GALAHAD.jl`, you can
 directly run `pkg> dev .`.
 
+## Documentation
+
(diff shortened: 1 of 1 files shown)


nimgould (approved):
Looks good to me. Does LBT support ATLAS BLAS? I only ask as this is often the best choice for people with exotic or new systems

amontoison:
Yes, it also supports ATLAS BLAS. If we don't have a JLL for that but you can provide your own path to your local BLAS / LAPACK and it works.
I did that with `liBHSL` for using NVBLAS on the Grace CPU Architecture of NVIDIA. We have two [GH200](https://www.jlse.anl.gov/nvidia-gh200) at Argonne.

PR #471, opened 2025-08-02
Move BNLS in forthcoming

amontoison:
BNLS is not `forthcoming` but I am not sure that it is production ready.


nimgould:
I have no idea why this is even part of the distro, it is simply a placeholder, and should be in forthcoming. I suggest you block it from meson. It will be a major task to finish this, and other things have overtaken it in my priorities

PR #477, opened 2025-08-04
Add a header galahad_callbacks.h

amontoison:
@nimgould I worked on a  `galahad_callbacks.h` to easily reuse the signatures of GALAHAD callbacks in the various packages (and ensure a uniform API).
I also find that it is easier to understand the signature of the callbacks in the headers that support direct communication.

The users don't need to write the prototypes in their code too, they can do something like this:
```shell
galahad_f fun;
galahad_g grad;
galahad_h hess;
```
instead of 
```shell
// Function prototypes
ipc_ fun(ipc_ n, const rpc_ x[], rpc_ *f, const void *);
ipc_ grad(ipc_ n, const rpc_ x[], rpc_ g[], const void *);
ipc_ hess(ipc_ n, ipc_ ne, const rpc_ x[], rpc_ hval[], const void *);
```
Which allow to ensure that their callbacks respect what the C interface of GALAHAD routines require.
Otherwise they will have a compilation error / warning.

I needed these new  "typedef" to have a clean signature in the multiprecision `galahad_c.h`.

### DIFF
--- a/GALAHAD.jl/gen/rewriter.jl
+++ b/GALAHAD.jl/gen/rewriter.jl
@@ -78,6 +78,14 @@ function rewrite!(path::String, name::String, optimized::Bool)
       end
     end
 
+    # Callbacks
+    for callback in ("galahad_f", "galahad_g", "galahad_h", "galahad_prec", "galahad_hprod",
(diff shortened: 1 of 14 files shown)


nimgould (approved):
This looks like a very good idea

PR #484, opened 2025-08-06
[CI] Add a workflow to check in we need to regenerate Julia wrappers

amontoison:


amontoison:
@nimgould @jfowkes 
I added a GitHub workflow that will inform us if we need to regenerate the Julia wrappers with all instructions:
Example: https://github.com/ralna/GALAHAD/actions/runs/16765273982/job/47469034946?pr=484#logs

It will be exactly the same thing for `galahad_c.h`.

PR #486, opened 2025-08-07
Remove ancient Python 2 GUI

jfowkes:
Remove the ancient Python 2.2 GUI for GALAHAD as Python 2 has been end-of-life since 2020.


nimgould (approved):
Cast it into the pool of eternal blackness

PR #489, opened 2025-08-09
[Meson] Compile the C tests of SSIDS

amontoison:
- Compile the C tests of SSIDS with Meson
- Add unit tests for SSIDS in Julia


amontoison:
@nimgould We have the following failure in single precision for `ssidst`:
```shell
================================== 424/449 ===================================
test:         GALAHAD:ssids+single+C / ssidstf_c_single
start time:   21:11:54
duration:     0.24s
result:       killed by signal 6 SIGABRT
command:      ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 LD_LIBRARY_PATH=/home/runner/work/GALAHAD/GALAHAD/builddir/:/opt/hostedtoolcache/Python/3.12.11/arm64/lib:/home/runner/work/GALAHAD/GALAHAD/galahad/lib:/home/runner/work/GALAHAD/GALAHAD/../deps/lib:/home/runner/work/GALAHAD/GALAHAD/../CUTEst/lib MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 MESON_TEST_ITERATION=1 MALLOC_PERTURB_=182 /home/runner/work/GALAHAD/GALAHAD/builddir/ssidstf_c_single
----------------------------------- stderr -----------------------------------
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
==============================================================================
```

nimgould:
I have spent the morning on this, and with no luck. 9 out of 10 times, the test runs correctly, but yes, the other 1 time, I see what you do (and it happens for 0 based indexing as well). gdb indicates something is
undefined deep in the C++ multi-threading, and I suspect the chance of me
finding this is zero. But it is interesting that this only seems to happen for the 64 bit integer case, which isn't part of the SPRAL version, so possibly something I missed.

nimgould:
I am afraid that we may have to live with this, I have run out of ideas. Might I suggest that we simply don't have the 64 bit C tests for ssids for the time being, unless one of you has a good idea how to track this down

amontoison:
It is fine for me to only run the C and Julia tests related to `SSIDS` with `Int32`.

PR #502, opened 2025-08-30
Galahad ssids

nimgould:
A full integration of the relevant parts of SPRAL, particularly ssids and the Rutherford-Boeing matrix utilities. These have been GALified, and in particular the ssids_options type is now ssids_control. The makefile versions have been tested, but I'll need actions to test the meson builds and runs.

PR #503, opened 2025-08-30
[SSIDS] Modifications of Alexis

amontoison:
- Use `galahad_precision.h` instead of `ssids_rit.h`
- Remove the C++ profiler of SSIDS

All modifications are on top of the branch `galahad_ssids` so it will not conflict with #502.


amontoison:
Yes, it will go in your branch.
I just merged it such that we only need to fix CI on your branch `galahad_ssids`.

PR #505, opened 2025-09-03
Don't use the macro d_ipc_ in C++ printf

amontoison:


amontoison:
This PR didn't fixed the issue but #506 yes.

PR #536, opened 2026-04-17
[WIP] Fix python memory leaks

### LINKED ISSUE #535: Memory leaks in Python interfaces
As reported for PyCUTEst in https://github.com/jfowkes/pycutest/issues/108, our usage of `Py_BuildValue` in the Python interfaces is incorrect.

Newly created NumPy arrays start with a reference count of 1:
```C
PyArrayObject *py_g = (PyArrayObject *) PyArray_SimpleNew(1, ndim, NPY_DOUBLE);
```
and are returned using the `Py_BuildValue` syntax:
```C
return Py_BuildValue("O", py_g);
```
however The `"O"` format specifier increments the reference count to 2:
> O (object) [PyObject *]
Pass a Python object untouched but create a new [strong reference](https://docs.python.org/3/glossary.html#term-strong-reference) to it (i.e. its reference count is incremented by one).

When the caller's tuple is eventually freed, the reference count drops to 1 but never reaches 0, so the array is never deallocated. This causes memory leaks. The correct format descriptor to use here is `"N"` which does not increase the reference count of the arrays:
> N (object) [PyObject *]
Same as O, except it doesn’t create a new [strong reference](https://docs.python.org/3/glossary.html#term-strong-reference).

Also there's occasionally some very suspect increasing of reference counts for `Py_BuildValue`:
```C
// Return x and g
bgo_solve_return = Py_BuildValue("OO", py_x, py_g);
Py_XINCREF(bgo_solve_return);
return bgo_solve_return;
```
since `Py_BuildValue` creates a new reference this is also clearly wrong. Not sure what I was thinking when I wrote this 🙈

jfowkes:
Work in progress to resolve #535  

!! not to be merged until the **beta_55** branch is merged !!

### DIFF
--- a/src/bgo/Python/bgo_pyiface.c
+++ b/src/bgo/Python/bgo_pyiface.c
@@ -448,7 +448,7 @@ static PyObject* py_bgo_initialize(PyObject *self){
 
     // Return options Python dictionary
     PyObject *py_options = bgo_make_options_dict(&control);
-    return Py_BuildValue("N, py_options);
+    return Py_BuildValue("N", py_options);
 
 }
 
--- a/src/bgo/Python/bgo_pyiface.c
+++ b/src/bgo/Python/bgo_pyiface.c
@@ -55,7 +55,7 @@ static int eval_f(int n, const double x[], double *f, const void *userdata){
     PyObject *py_x = PyArray_SimpleNewFromData(1, xdim, NPY_DOUBLE, (void *) x);
 
     // Build Python argument list
-    PyObject *arglist = Py_BuildValue("(O)", py_x);
+    PyObject *arglist = Py_BuildValue("(N)", py_x);
 
     // Call Python eval_f
     PyObject *result = PyObject_CallObject(py_eval_f, arglist);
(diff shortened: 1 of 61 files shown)


src/bgo/Python/bgo_pyiface.c
```diff
@@ -449,7 +448,7 @@ static PyObject* py_bgo_initialize(PyObject *self){
 
     // Return options Python dictionary
     PyObject *py_options = bgo_make_options_dict(&control);
-    return Py_BuildValue("O", py_options);
+    return Py_BuildValue("N, py_options);
```
amontoison:
```suggestion
    return Py_BuildValue("N", py_options);
```

jfowkes: Good catch, many thanks!

jfowkes:
Nick is away on leave until May so we won't be able to fix the SLLS 64bit build errors on the beta_55 branch until he gets back, should be an easy enough fix though:
```
./src/slls/slls.F90:1797:65:

 1797 |               bad_alloc = inform%bad_alloc, out = control%error )
      |                                                                 1
Error: There is no specific subroutine for the generic ‘space_resize_array’ at (1)
```

nimgould:
Thank you, merged into beta_55. I've just got back, and it may take me  a day or two to get back onto the horse :)

Issue #269, opened 2024-02-22
Name of the libraries when GALAHAD is compiled with 64-bit integer


amontoison:
We should decide what is the name of the static / shared library when we use 64-bit integer.
With 32-bit integer, we generate `libgalahad_single` and `libgalahad_double`.
I suggest `libgalahad64_single` and `libgalahad64_double`.

cc @nimgould

nimgould:
elsewhere we have reversed things so that packages are _single_64 not 64_single

amontoison:
Ok, so we should go with `libgalahad_double_64` or `libgalahad_double64.so`.
For libHSL, I generate a `libhsl64.so`.

Issue #15, opened 2022-11-05
Undefined symbols -- shared libraries


amontoison:
@nimgould
I almost fixed all links and undefined symbols when we generate the shared libraries [here](https://github.com/ralna/GALAHAD/blob/master/bin/create_shared#L46-L58).
```shell
gcc -shared -o shared/libgalahad_dummy.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_dummy.a -Wl,--no-whole-archive

gcc -shared -o shared/libgalahad_metis.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_metis.a -Wl,--no-whole-archive

gcc -shared -o shared/libgalahad_cutest_dummy.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_cutest_dummy.a -Wl,--no-whole-archive

gcc -shared -o shared/libgalahad_umfpack.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_umfpack.a -Wl,--no-whole-archive

gcc -shared -o shared/libgalahad_mumps.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_mumps.a -Wl,--no-whole-archive

gcc -shared -o shared/libgalahad_pastix.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_pastix.a -Wl,--no-whole-archive

gcc -shared -o shared/libgalahad_wsmp.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_wsmp.a -Wl,--no-whole-archive

gcc -shared -o shared/libgalahad_pardiso.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_pardiso.a -Wl,--no-whole-archive

gcc -shared -o shared/libgalahad_mkl_pardiso.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_mkl_pardiso.a -Wl,--no-whole-archive

gcc -shared -o shared/libgalahad_minpack.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_minpack.a -Wl,--no-whole-archive -lgfortran

gcc -shared -o shared/libgalahad_hsl.so -Wl,--no-undefined \
-Wl,--whole-archive double/libgalahad_hsl.a -Wl,--no-whole-archive -lgfortran

[...]

Issue #14, opened 2022-11-02
Segmentation fault with LANCELOT B tests


amontoison:
@nimgould 
```shell
alexis@HP-Spectre:~/Bureau/git/GALAHAD/src$ make -s -f $GALAHAD/makefiles/$arch test
 Compiling lancelott      		[ OK ]

 Exhaustive test of subroutine interface to LANCELOT B

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7f139b87bad0 in ???
#1  0x7f139b87ac35 in ???
#2  0x7f139b58b51f in ???
	at ./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0
#3  0x5601e5f0835c in ???
#4  0x5601e5eb5bbe in ???
#5  0x5601e5ece111 in ???
#6  0x5601e5ea6350 in ???
#7  0x5601e5eaa7d3 in ???
#8  0x7f139b572d8f in __libc_start_call_main
	at ../sysdeps/nptl/libc_start_call_main.h:58
#9  0x7f139b572e3f in __libc_start_main_impl
	at ../csu/libc-start.c:392
#10  0x5601e5ea22f4 in ???
#11  0xffffffffffffffff in ???
Segmentation fault (core dumped)
 Tests for LANCELOT B complete
```

nimgould:
I don't see this here. Which platform? Could you recompile with e.g, DEBUG = -g -fbacktrace
in $GALAHAD/makefiles/(architecture) for your architecture, i.e.
cd $GALAHAD/src/lancelot
make -s -f (architecrure) cleanall
make -s -f (architecrure) test
and report back. Thank you

amontoison:
My platform is Linux with gcc / gfortran compilers.
```shell
Backtrace for this error:
#0  0x7f962a11dad0 in ???
#1  0x7f962a11cc35 in ???
#2  0x7f9629e2d51f in ???
	at ./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0
#3  0x55e08ab4635c in __lancelot_precn_double_MOD_precn_use_preconditioner
	at /home/alexis/Bureau/git/GALAHAD/objects/pc64.lnx.gfo/double/precn.f90:931
#4  0x55e08aaf3bbe in lancelot_solve_main
[...]

Issue #13, opened 2022-11-01
Issues with pastix


amontoison:
@nimgould 
If I remove the commit 54a99dec19377758f2a2112ba24401632dcb8db5, I have this error:

```shell
 Installing the double precision version
 compiling in /home/alexis/Bureau/git/GALAHAD/src with the comand
 /usr/bin/make -s -f /home/alexis/Bureau/git/GALAHAD/makefiles/pc64.lnx.gfo all_cutest
 Compiling cpu_time       		[ OK ]
 Compiling cutest_dummy   		[ OK ]
 Compiling spm_interfaces  		[ OK ]
make[5]: ***  Aucune règle pour fabriquer la cible « ../../dum/spmf.f90 », nécessaire pour « /home/alexis/Bureau/git/GALAHAD/objects/pc64.lnx.gfo/double/libgalahad_pastix.a(spmf.o) ». Arrêt.
make[4]: *** [/home/alexis/Bureau/git/GALAHAD/src/basic/makemaster:324 : pastix_silent] Erreur 2
make[3]: *** [/home/alexis/Bureau/git/GALAHAD/src/makedefs/intermediaries:161 : make_dummy] Erreur 2
make[2]: *** [/home/alexis/Bureau/git/GALAHAD/src/lancelot/../general/makemaster:629 : zd11] Erreur 2
make[1]: *** [/home/alexis/Bureau/git/GALAHAD/src/lancelot/makemaster:338 : basics] Erreur 2
make: *** [/home/alexis/Bureau/git/GALAHAD/src/makemaster:43 : all_cutest_double] Erreur 2
```
I used the options described in #12.

nimgould:
Missing dummy spmf.f90 and pastixf.f90 pushed.

PR #542, opened 2026-08-03
Replace C++ code of SSIDS with Fortran

amontoison:
Remove all the C++ and GPU related code of SSIDS.
I started that a long time ago...
It also removes the need of `hwloc`.


nimgould:
Thanks, Alexis, that is really great. I will have a look as soon as I can ... I have managed to get a short-term temporary licence for the NAG library, and need to give this (the interface from the new QP solver to NAG) priority before the licence expires next week. I see that you also updated the makemaster file ... thank you!!

I do actually wonder if we should rename the package so as not to confuse the world, particularly as Hussam is planning for the official ssids to evolve. And I need to write docs, and a Python interface.

Do you use AI to help you here?

jfowkes:
Fantastic thanks @amontoison! We should also incorporate Hussam's adaptive front size fix in  https://github.com/ralna/spral/pull/271 as this results in a major performance improvement.   

Agreed we should probably change the name as the codebase has diverged so much at this point, unless @haldaas is interested in merging this upstream.

nimgould:
Sorry to be an idiot, as usual

I did a git checkout -b spike-ssids-cholesky-fortran

which seems to have set things here, 

git branch --show-current
spike-ssids-cholesky-fortran

but when I tried a git pull, I get 

here is no tracking information for the current branch.
Please specify which branch you want to rebase against.
See git-pull(1) for details.

    git pull <remote> <branch>

If you wish to set tracking information for this branch you can do so with:

    git branch --set-upstream-to=origin/<branch> spike-ssids-cholesky-fortran

Should I follow this advice?

jfowkes:
Assuming that you did the checkout from the master branch that should work, as you created a new local branch off of the branch you were on previously.

More standard practice is to do a `git fetch` to get the names of any new remote branches and then a `git checkout branchname`.

nimgould:
On the makefile build under gfortran I see:

Warning: Unused variable ‘adc’ ssids_factor.F90:720:35:
Warning: Unused variable ‘adr’ ssids_factor.F90:725:29:
Warning: Unused variable ‘adum’ ssids_factor.F90:719:35:
Warning: Unused variable ‘blk’
Warning: Unused variable ‘ii’ ssids_factor.F90:726:43:
Warning: Unused variable ‘jj’ ssids_factor.F90:719:75:
Warning: Unused variable ‘k’ ssids_factor.F90:719:66:
Warning: Unused variable ‘nr’ ssids_factor.F90:729:60:
Warning: Unused variable ‘tbp’ ssids_factor.F90:730:24:
Warning: Unused variable ‘td11’ ssids_factor.F90:730:30:
Warning: Unused variable ‘td21’ ssids_factor.F90:729:35:
Warning: Unused variable ‘tdoff’ ssids_factor.F90:731:23:
Warning: Unused variable ‘tfin’ ssids_factor.F90:729:55:
Warning: Unused variable ‘tk’ ssids_factor.F90:732:39:
Warning: Unused variable ‘tldw’ ssids_factor.F90:729:51:
Warning: Unused variable ‘tlflag’ ssids_factor.F90:726:70:
Warning: Unused variable ‘tmpi’ ssids_factor.F90:729:23:
Warning: Unused variable ‘tnc’ ssids_factor.F90:729:43:
Warning: Unused variable ‘tnelim’ ssids_factor.F90:729:65:
Warning: Unused variable ‘tnp’ ssids_factor.F90:729:28:
Warning: Unused variable ‘tnr’ ssids_factor.F90:729:28:

nothing serious of course. But when I run the test program 
ssidst.F90, I get 

Exhaustive test of subroutine interface to ssids
At line 330 of file ssids_factor.F90
Fortran runtime error: Index '6' of dimension 1 of array 'a' above upper bound of 5

Error termination. Backtrace:
#0  0x13185e623e59 in ???
#1  0x13185e624a71 in ???
#2  0x13185e625082 in ???
#3  0x626c1e5975f7 in __galahad_ssids_factor_double_MOD_ldlt_tpp_factor
	at /home/nimg/Dropbox/fortran/optrove/galahad/objects/pc64.lnx.gfo/double/ssids_factor.F90:330
#4  0x626c1e59bd7c in __galahad_ssids_factor_double_MOD_run_pivoted._omp_fn.0
	at /home/nimg/Dropbox/fortran/optrove/galahad/objects/pc64.lnx.gfo/double/ssids_factor.F90:923
#5  0x13185e3e65bb in ???
#6  0x626c1e58f3f5 in run_pivoted
	at /home/nimg/Dropbox/fortran/optrove/galahad/objects/pc64.lnx.gfo/double/ssids_factor.F90:898
#7  0x626c1e597fb9 in __galahad_ssids_factor_double_MOD_ldlt_app_factor
	at /home/nimg/Dropbox/fortran/optrove/galahad/objects/pc64.lnx.gfo/double/ssids_factor.F90:787
#8  0x626c1e59bb2e in __galahad_ssids_factor_double_MOD_run_pivoted._omp_fn.0
	at /home/nimg/Dropbox/fortran/optrove/galahad/objects/pc64.lnx.gfo/double/ssids_factor.F90:911
#9  0x13185e3e40d2 in ???
#10  0x13185e3ed937 in ???
#11  0x626c1e58d674 in __galahad_ssids_factor_double_MOD_factor_subtree_delay._omp_fn.0
	at /home/nimg/Dropbox/fortran/optrove/galahad/objects/pc64.lnx.gfo/double/ssids_factor.F90:2143
#12  0x13185e3ea6fd in ???
#13  0x13185e09cb83 in start_thread
	at ./nptl/pthread_create.c:447
#14  0x13185e129d6b in clone3
	at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
#15  0xffffffffffffffff in ???
  ssids double precision tests

in a call to GEMM. Superficially, it looks like this is trying to  access a(m+1,*) of an a(m,n) matrix. So either the argument a(nelim+2, nelim+1) is wrong or the dimensions m-nelim-1, n-nelim-1 are? (perhaps a 0-based mistake)?

Looking at the original C++, I think it should be a(nelim+1,nelin+1), but that could be wrong ... those pesky 0-based flattened index arrays strike again!! I wonder, going forward, if we shouldn't try to make everything here 1-based? It is not very natural in fortran to have nelim going from 0 to n-1 ...

My only other comment here is that it would be nice to put back the C comments (albeit as fortran ones) as these do help explain what the code is trying to do. No hurry on this

nimgould:
With the NAG compiler, I get 

Compiling types          		Error: types.F90, line 400: No data-edit descriptor for effective item
Error: types.F90, line 401: No data-edit descriptor for effective item
Error: types.F90, line 402: No data-edit descriptor for effective item
Error: types.F90, line 403: No data-edit descriptor for effective item
Error: types.F90, line 404: No data-edit descriptor for effective item
Error: types.F90, line 405: No data-edit descriptor for effective item
Error: types.F90, line 431: No data-edit descriptor for effective item
Error: types.F90, line 433: No data-edit descriptor for effective item
Error: types.F90, line 446: No data-edit descriptor for effective item
Error: types.F90, line 448: No data-edit descriptor for effective item
which seems to suggest that NAG doesn't like the    CLASS( SSIDS_control_type ), INTENT( IN ) :: this

jfowkes:
Aren't "No data-edit descriptor for effective item" errors to do with incorrect format descriptors in write statements? Should hopefully be an easy fix.

nimgould:
Yes, indeed, the error is the 200 FORMAT( '( A, I15 )' )
 which should, of course, be 200 FORMAT( A, I15 ) (no quotes)

nimgould:
Same in the following subroutine

200 FORMAT(  //, 3A, I2, A  )
210 FORMAT(  // A, 5( / A ,I12  ), 5( / A, ES12.4 )  )

OK, fixed those, and removed the VOLATILE from
 TYPE( contrib_type ), POINTER, VOLATILE :: fcontrib 
in contrib_iface.F90 (as AI says that it isn't necessary), the code now compiles under nagfor (a first!!), but then falls over at exactly the same place as for gfortran (above). I'll see what I can do ...

nimgould:
Needless to say, nothing worked with ifx, when compiling ssids.F90 I get an error #5633: **Internal compiler error: segmentation violation signal raised** Please report this error along with the circumstances in which it occurred in a Software Problem Report.  Note: File and line given may not be explicit cause of this error.
compilation aborted for ssids.F90 (code 3)

with no clues to why it failed (deep in the libc stuff). But this is for oneapi/compiler/2024.2, I haven't got a more modern one, I'll see if I can get one ...

OK, it compiles fine under 2026.1, but the test fails with the uninformative  symbol lookup error: ../ssids/run_ssids: undefined symbol: __kmpc_alloc, version VERSION

nimgould:
Finally, it compiles and runs with nvfortran

amontoison:
> OK, it compiles fine under 2026.1, but the test fails with the uninformative symbol lookup error: ../ssids/run_ssids: undefined symbol: __kmpc_alloc, version VERSION

You need to link with the OpenMP library of Intel: `libiomp5.so`.
it is why you have undefined symbols.
Don't hesitate to push directly on this branch if you fix something.
I can't work on it before this weekend.

nimgould:
I tried linking against /opt/intel/oneapi/compiler/latest/lib/libiomp5.so
but nothing changed. I have compiled and linked with -fopenmp, I presume that is ok?

Update: it was my fault, flang also has a libiomp5.so, and was further up my link chain. Changing the order sorted things out.

jfowkes:
I believe that to link against Intel OpenMP you need to use `-fiopenmp`

nimgould:
I haven't tried that, but I will ... that seems to work too. I still get the very annoying "OMP: Info #269: OMP_NESTED variable deprecated, please use OMP_MAX_ACTIVE_LEVELS instead" message, even though I have set OMP_MAX_ACTIVE_LEVELS ... but it seems that I actively have to remove OMP_NESTED to get rid of the message, doh!

nimgould:
OK, I'm making progress.

As I suspected, the call to gemm in  ldlt_tpp_factor has an incorrect leading array index for both arguments a(.,), these
should be (I believe) nelim+1 not nelim+2. This allows gfortran to run to completion. But, fortunately, nagfor is a stricter compiler, and objects that the second "a" argument a(nelim+1, nelim+2) lies outside the array bounds. gemm will not complain as the m, n argumemts are 0, and gemm will no-op for this. To avoid this,
simply adding a IF ( nelim < n - 1 ) & before the CALL DGEMM does the job.

The next issue is that ldlt_app_solve_diag fails for an edge case;
there is an
 
IF (i+1 == n .OR. IEEE_IS_FINITE(d(2*i+3))) THEN ...

the first clause is true, but fortran doesn't say in which order clauses should be checked, and the index for d in the 2nd is outside bounds. So a 
is1x1 = ( i+1 == fn%nelim )
            IF ( .NOT. is1x1 ) is1x1 = ieee_is_finite_local( fn%d( 2*i+3 ) )
solves the issue (I then saw that you did something almost identical in SUBROUTINE alter in numerical_subtree.F90 ;)

But I am now stuck on the alter test. In SUBROUTINE alter (the one in ssids.F90 not numerical_subtree.F90 ) calls subtree%alter() with a segment of the 2-D array that stores the B block (dimension 1 is for the diagonal and dimension 2 for the off diagonal). There is a single part to this array (akeep%nparts = 1), but the 
d( 1 : 2, akeep%part( part ) :                                        akeep%part( part + 1 ) - 1 )
gives a d(1:2,1:1) array. It should be d(1:2,1:5). I don't know why
this is the case, I don't know where akeep%part is set.  In any event, it causes an out-of-range exception in subtree%alter() as the array is the wrong shape.

Anyway, I'll check all I have so far in, but will be off line for three days ... we have visitors AND my home ethernet is boosting from 32Mbps to 900, as total fibre has finally arrived in the village ... and I very much suspect that my internet provider will mess it up!

nimgould:
I have pushed an update that I believe fixes the nagfor issue. This is in subroutine alter in ssids, in the call to subtree%alter. I think that this should be 
  CALL subtree%alter( d( 1, akeep%part( part ) ) )
not
  CALL subtree%alter( d( 1 : 2, akeep%part( part ) :                                            akeep%part( part + 1 ) - 1 ) )
as the latter tries to pass a 2x1 vector, while I suspect we should instead be passing a pointer to the first entry in a 2xn matrix (I think that is what the origial C++ code is doing). In any event, the simple ssids test now passes with all my compilers with the makefile install.

Given this, I am happy if you merge, but perhaps we should change ssids -> slblt everywhere first?

nimgould:
If we are to take control of this package, really we should improve the test example as it is very primitive at the moment

amontoison:
> I believe that to link against Intel OpenMP you need to use `-fiopenmp`

@nimgould @jfowkes 
https://github.com/ralna/GALAHAD/blob/master/meson.build#L165-L176

`-fopenmp` should also work for `ifort` / `ifx` but the documented flag is `-qopenmp` on linux / mac and `/Qopenmp` on windows.

amontoison:
@nimgould I am working on the renaming `ssids` -> `slblt` everywhere.
I agree to update the test / example but let's merge this PR before such that it is easier to review. 

Update: Nick can you do a final checking and merge the PR (squash + merge) please ?

nimgould:
Agreed. Tests for slblt pass here

PR #543, opened 2026-08-08
Fix HSL dummy module ABI

amontoison:
My colleague Paulo Silva (Brazil) wants to build `Algencan` against the public dummy HSL Fortran modules and then hot-swap the licensed HSL library at run time (the `HSL_jll` scenario).

In this setup, Algencan is compiled against the dummy .mod files that we have here (and in the dummy libHSL repository), but at run time the real, licensed `libhsl_subset.so` is loaded in place of the dummy.

For that swap to be safe, every public derived type in the dummy module must be ABI-identical to the licensed one, same storage_size, same byte offset for every public component, across all real precisions (single/double/quadruple) and both integer sizes (int32/int64).

A Fortran .mod file is a compile-time contract: the caller bakes in the type's memory layout.
If the dummy's layout differs from the licensed library's (a field in the wrong order, the wrong integer kind (ip_ vs long_), a missing/extra field, or a handle of the wrong size), then at run time the licensed routine reads and writes at different offsets than the caller expects.
Nothing warns you: it just corrupts memory, returns garbage, or crashes.
That is exactly what was happening.

Concretely, several dummy types were not faithful to the licensed layout:
  - MA86_info had its num_* counters in the wrong order;
  - the opaque handles ma86_keep, ma87_keep, ma97_akeep/fkeep had the wrong total size;
  - ma48_factors/control/ainfo/finfo were missing long_ fields and used the wrong integer kinds;
  - ma97_control/ma97_info and mi35_control were missing/mis-ordered fields.

Because the layout was wrong, the hot-swap silently broke, and this is the same mechanism behind the intermittent failures I was still seeing from Julia with `GALAHAD` (via `GALAHAD_jll`, which relies on the very same dummy modules).

After this PR, the dummy modules are a faithful stand-in for the licensed library at both compile time and run time, so Algencan (and GALAHAD from Julia) can hot-swap the real HSL solvers safely.
We should be good.

cc @nimgould @jfowkes

### DIFF
--- a/src/external/hsl/hsl_ma48/hsl_ma48r.f90
+++ b/src/external/hsl/hsl_ma48/hsl_ma48r.f90
@@ -54,17 +54,17 @@ module hsl_ma48_real
    end interface
 
    type ma48_factors
-     private
-      integer(ip_),  allocatable :: keep(:)
-      integer(ip_),  allocatable :: irn(:)
-      integer(ip_),  allocatable :: jcn(:)
+      integer(long_), allocatable :: keep(:)
+      integer(long_), allocatable :: irn(:)
+      integer(long_), allocatable :: jcn(:)
       real(rp_), allocatable :: val(:)
       integer(ip_) :: m       ! Number of rows in matrix
       integer(ip_) :: n       ! Number of columns in matrix
-      integer(ip_) :: lareq   ! Size for further factorization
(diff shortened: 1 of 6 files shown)


nimgould:
Thanks, Alexis. This makes sense. But note that fortran does not impose a storage sequence order on components unless the SEQUENCE attribute is specified.

PR #547, opened 2026-08-15
[CI] Run Python tests on Windows

amontoison:


jfowkes:
@amontoison the int32 tests are still all failing :(

jfowkes:
Note that Python always uses int64, no matter what we use in the backend.

jfowkes:
@amontoison since we only cross-compile the Python interfaces with galahad_double it probably doesn't make sense to test them with galahad_single?

Here are the current cross-compilation options from `pyproject.toml`:
```toml
[tool.meson-python.args]
setup = ['-Dpythoniface=true', '-Dpython.install_env=auto', '-Dsingle=false', '-Ddouble=true', '-Dquadruple=false', '-Dtests=false']
install = ['--tags=runtime,python-runtime']
```

amontoison:
Sorry @jfowkes but I don't understand any of your comments.
If you can do longer sentences with more context, it will help :)

> @amontoison the int32 tests are still all failing :(

Yes, we only compile the Python interface with `libgalahad_double.$dlext` (Int32 + Float64) but I think the issue is related to the environment (Windows runtime different than the one used for Python).

> Note that Python always uses int64, no matter what we use in the backend.

Python is using Int64 as default integer or for indexing of arrays, or both?
Why are we using `NPY_LONG` in `galahad_python.h` in that case?
To the best of my knowledge, `NPY_LONG` in int32 on Windows.

>@amontoison since we only cross-compile the Python interfaces with galahad_double it probably doesn't make sense to test them with galahad_single?

> Here are the current cross-compilation options from `pyproject.toml`:
```toml
[tool.meson-python.args]
setup = ['-Dpythoniface=true', '-Dpython.install_env=auto', '-Dsingle=false', '-Ddouble=true', '-Dquadruple=false', '-Dtests=false']
install = ['--tags=runtime,python-runtime']
```
I don't understand at all the goal of this comment.
We only support `libgalahad_double` for the Python interfaces so we cross-compile only this library with `cibuildwhells`.
But here it is the CI of GALAHAD, and we compile and run the tests of C, Fortran, Python and Julia. If we compile also `libgalahad_single.$dlext` and `libgalahad_quadruple.$dlext`, it will not used for Python but still used for the other languages. 

In all cases I don't see the relation with your comment and the current PR.
It could make sense if we add dedicated CI jobs that test the cross-compilation on `cibuildwheels` + Python tests of GALAHAD.
We have that in Uno: https://github.com/cvanaret/Uno/blob/main/.github/workflows/build-python-wheels.yml

jfowkes:
Sorry @amontoison I will try and write longer sentences with more context :) 

> Python is using Int64 as default integer or for indexing of arrays, or both?
Why are we using NPY_LONG in galahad_python.h in that case?
To the best of my knowledge, NPY_LONG in int32 on Windows.

Yes Python's default integer is 64bit and since it also uses the default integer for indexing arrays, array indices are also 64bit (it is possible for the user to construct 32bit value arrays, but we don't support these and such arrays would still have 64bit indices). I have no idea why we are using `NPY_LONG` in `galahad_python.h`, looking at the git history it seems @nimgould changed it from `NPY_INT64` to `NPY_LONG` two years ago for some reason. Changing it back to `NPY_INT64` should hopefully fix things or do we need to do more work to support Windows?

> I don't understand at all the goal of this comment.
We only support libgalahad_double for the Python interfaces so we cross-compile only this library with cibuildwhells.
But here it is the CI of GALAHAD, and we compile and run the tests of C, Fortran, Python and Julia. If we compile also libgalahad_single.$dlext and libgalahad_quadruple.$dlext, it will not used for Python but still used for the other languages.

Sorry I got confused between galahad_double/galahad_single and 64bit integer support within GALAHAD. At the moment the CI is running Windows Python tests for GALAHAD compiled with 32bit integers and it's failing. The question is should we be cross-compiling GALAHAD with 32bit integers for use with Python or should we instead enable `-Dint64=true` in `pyproject.toml` and build GALAHAD with 64bit integers for Python users instead?

nimgould:
Could have been me, no idea why I would do this though, I know less about Python than Donald J. Trump does

amontoison:
Hello @nimgould and @jfowkes,

I suggest that we revert the modification and go back to `NPY_INT64`, it is more portable between platforms.

I think it also makes sense to use `libgalahad_double_64.$dlext` from what I understand.
We will be able to pass directly the integer arrays of Python without conversions.
We never test it with CI but it is two lines to change.

I don't know why we didn't do that before but it may be because we didn't supported `-Dint64` yet.
We released GALAHAD 5 without it for sure.

Jari, do you have bandwidth to have a look?
I can also work on it in two weeks.

nimgould:
Agreed. Thanks, @amontoison . I'll be in and out for the next five weeks ... its my vacation month :)

jfowkes:
@amontoison sure happy to take a look. I think we need to revert your commit first though?

amontoison:
@jfowkes I removed the last two commits.

jfowkes:
Thanks @amontoison, should have time to fix this tomorrow.

jfowkes:
@amontoison looking into this in more detail it seems that NumPy actually uses LONG for integers by default 🙈 
https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.int_
so Nick was correct to have changed it from `NPY_INT64` to `NPY_LONG`.

This then brings us back to the original question, why do the tests break on Windows?

amontoison:
We got two issues, one is related to the Microsoft runtime that is different between what is installed here in CI and the one in Python but this problem is fixed when we cross-compile the wheels.
The second issue is for us, because `long` is `c_int` on Windows, some cast / conversion and failing in the Python tests.

jfowkes:
The cast conversions are going to be very difficult for me to debug as I don't have a Windows machine.

amontoison:
You can easily ssh a Windows runner if you uncomment these lines:
https://github.com/ralna/GALAHAD/blob/master/.github/workflows/meson_build.yml#L563

jfowkes:
Okay I will try to debug the failing `Windows/Int32/build` with gdb using SSH.

jfowkes:
I cannot debug it with SSH, it just hangs:
```
Run mxschmitt/action-tmate@v3
resolving dependencies...
looking for conflicting packages...
Packages (4) libevent-2.1.12-4  libssh-0.12.0-1  msgpack-c-6.1.0-2  tmate-2.4.0-5
Total Download Size:   0.55 MiB
Total Installed Size:  1.63 MiB
:: Proceed with installation? [Y/n] 
:: Retrieving packages...
 libssh-0.12.0-1-x86_64 downloading...
 tmate-2.4.0-5-x86_64 downloading...
 libevent-2.1.12-4-x86_64 downloading...
 msgpack-c-6.1.0-2-x86_64 downloading...
checking keyring...
checking package integrity...
loading package files...
checking for file conflicts...
:: Processing package changes...
installing libevent...
Optional dependencies for libevent
    python: to use event_rpcgen.py
installing libssh...
installing msgpack-c...
installing tmate...
```

jfowkes:
Okay I think I've finally got to the bottom of this, it seems that the NumPy `int` type depends entirely on which version of NumPy is installed, hence my confusion:

**NumPy 1.x (Legacy)**
In NumPy 1.x, the default integer type on Windows (even on 64-bit systems) is `long` (a 32-bit signed integer). 
- NumPy Dtype: `np.int32` (aliased as `np.int_`)
- C Backing Type: `long`

**NumPy 2.0 and newer (Released 2024+)**
In NumPy 2.0+, the default integer type on 64-bit Windows is `long long` (a 64-bit signed integer). 
- NumPy Dtype: `np.int64` (aliased as `np.intp`)
- C Backing Type: `long long` (or `int64_t`)

and annoyingly we of course have to support both versions of NumPy 🙈

jfowkes:
Right fixed the cast conversion issues, now we're seeing on every Windows Python package:
```
exit status 3221225477 or 0xc0000005
```
which is a Windows access violation error. I assume this is what you were referring to @amontoison about the different Microsoft runtimes?

amontoison:
> Right fixed the cast conversion issues, now we're seeing on every Windows Python package:
> 
> ```
> exit status 3221225477 or 0xc0000005
> ```
> 
> which is a Windows access violation error. I assume this is what you were referring to @amontoison about the different Microsoft runtimes?

Yes, from what I understand it is an incompatibility between the toolchain used for CI and the one by Python. :(

Is it possible to use the script of `cibuildwheels` as CI tests just to check compilation / unit tests?
Right now we only use when we trigger a new release.
I think it works well on Mac and Linux, I am not confident for the Windows platform.

jfowkes:
> Is it possible to use the script of `cibuildwheels` as CI tests just to check compilation / unit tests? Right now we only use when we trigger a new release. I think it works well on Mac and Linux, I am not confident for the Windows platform.

Yes absolutely, the only issue is that our Python tests are just a bunch of scripts so I'll have to add something like:
```
python test_arc.py && python test_bgo.py && python test_ugo.py...
```
I will see if I can add something like this to the `build_wheels.yml` GitHub action.

jfowkes:
My AI friend reckons it's a side-effect of the way the Windows GitHub-actions runners are setup and that

> The cleanest way to compile mixed C/Fortran code for modern Python on Windows is using the UCRT64 environment provided by MSYS2. This ensures your compiler produces binaries completely compatible with NumPy's runtime. Modify your GitHub Actions YAML to isolate the build to MSYS2's UCRT64:
> ```
> - name: Set up MSYS2 with UCRT64 Toolchain
>   uses: msys2/setup-msys2@v2
>   with:
>     msystem: UCRT64
>     update: true
>     install: >-
>       mingw-w64-ucrt-x86_64-gcc
>       mingw-w64-ucrt-x86_64-gcc-fortran
>       mingw-w64-ucrt-x86_64-ninja
>       mingw-w64-ucrt-x86_64-meson
>       mingw-w64-ucrt-x86_64-python
>       mingw-w64-ucrt-x86_64-python-numpy
> 
> - name: Build with Meson inside MSYS2
>   shell: msys2 {0}
>   run: |
>     meson setup builddir
>     meson compile -C builddir
> ```

Not that I know what any of this means, but it's probably worth trying... lots of fun for next week 😃

nimgould:
Thanks for trying to sort this out, chaps. The comination of Windows and Python would have sent me straight to an asylum. Have a good weekend

jfowkes:
Okay I've switched the Windows build to using MSYS2 with UCRT64 (the new Universal C Runtime `ucrtbase.dll`) rather than MYSYS2 with MINGW64 (the old `msvcrt.dll`). The Python tests now mostly work 🎉 🥳 

However, we do have still have two failures with exit status 1:
- fdc+Python - GALAHAD:fdc_py 
- sha+Python - GALAHAD:sha_py

which look like genuine bugs in the Windows Python interface. Fair enough.

More annoyingly, there are still a few failures with the dreaded exit status 3221225477 (Access Violation):
- bgo+Python - GALAHAD:bgo_quadratic_py
- bsc+Python - GALAHAD:bsc_py
- cro+Python - GALAHAD:cro_py
- expo+Python - GALAHAD:expo_py
- nls+Python - GALAHAD:nls_py

which are more puzzling. Are these genuine segfaults? @nimgould is there an underlying dependency or data structure that the above packages share? Seems like a random selection to me... I could really do with a Windows machine.

As an unfortunate side-effect this has broken the Julia, which now throws an access violation 🙈 @amontoison I guess we'll need to move out the Julia test on Windows as Julia seems to use the MINGW64 runtime whereas Python uses UCRT64.

amontoison:
> Okay I've switched the Windows build to using MSYS2 with UCRT64 (the new Universal C Runtime `ucrtbase.dll`) rather than MYSYS2 with MINGW64 (the old `msvcrt.dll`). The Python tests now mostly work 🎉 🥳
> 
> However, we do have still have two failures with exit status 1:
> 
> * fdc+Python - GALAHAD:fdc_py
> * sha+Python - GALAHAD:sha_py
> 
> which look like genuine bugs in the Windows Python interface. Fair enough.
> 
> More annoyingly, there are still a few failures with the dreaded exit status 3221225477 (Access Violation):
> 
> * bgo+Python - GALAHAD:bgo_quadratic_py
> * bsc+Python - GALAHAD:bsc_py
> * cro+Python - GALAHAD:cro_py
> * expo+Python - GALAHAD:expo_py
> * nls+Python - GALAHAD:nls_py
> 
> which are more puzzling. Are these genuine segfaults? @nimgould is there an underlying dependency or data structure that the above packages share? Seems like a random selection to me... I could really do with a Windows machine.
> 
> As an unfortunate side-effect this has broken the Julia, which now throws an access violation 🙈 @amontoison I guess we'll need to move out the Julia test on Windows as Julia seems to use the MINGW64 runtime whereas Python uses UCRT64.

Excellent work @jfowkes ! 🤘 
We need a runtime "matrix" in the CI for Windows (MINGW64 / UCRT64).
Depending on the value of the runtime, we can easily only run the Python OR Julia tests.
I will have a look next weekend.

I can also wait that you fix the remaining issues in the Python interfaces on Windows such that I can enable again all the CI jobs.

meson.build
```diff
@@ -824,7 +824,7 @@ if build_tests and build_ciface
 endif
 
 # Python tests
-if build_tests and build_pythoniface and build_ciface and build_double and (not int64) and (host_system != 'windows')
+if build_tests and build_pythoniface and build_ciface and build_double and (not int64)
```
amontoison: @jfowkes Should we drop the constraint `(not int64)`?
From what I understand, it should also work if we use `libgalahad_double_64.$dlext`.

nimgould:
@jfowkes , there is a link bsc -> nls and bsc -> expo (and thse are the only exposed packages that use bsc) , but other than that, nothing obvious.

nimgould:
Sorry, no idea how I closed this ...

meson.build
jfowkes: @amontoison yeah I think we should. Also should we build the precompiled Python "wheels" with int64 since NumPy on all platforms now has 64bit integers?

jfowkes:
No worries, I will do some digging when I have time and see if I can get to the bottom of the failing packages.

meson.build
jfowkes: @amontoison I tried that but it breaks horribly at the linking stage if I drop the constraint `(not int64)`

amontoison: Oh it is because all symbols have the suffix `_64`. We don't use the preprocessing macros in the Python interfaces.

jfowkes: Ah okay, that's probably best left for another PR then.

amontoison: I took 30 minutes tonight to support the 64-bits integer C API in the Python interfaces.
We need to replace `int` by `ipc_` now but I don't know how to handle the various `Long int` that we have a little bit everywhere in the Python interfaces.

@jfowkes What should we do? Should we support both int32 / int64 libraries of GALAHAD (`libgalahad_double.$dlext` / `libgalahad_double_64.$dlext`) and do some cast everywhere to support both `numpy` integers?

I think we should expose only one API for Python users that is independent of all these integer variants and handle all the conversions internally.
99.9% of the users will install / use it with PyPI.

jfowkes: @amontoison Agreed. There doesn't seem any point to me in supporting for Python both int32 / int64 libraries of GALAHAD (libgalahad_double.$dlext / libgalahad_double_64.$dlext). Given that modern NumPy on all platforms is 64bit integer, I would say that the correct thing to do is to only build the Python interfaces with the int64 library of GALAHAD and then we don't need to worry about any cast conversions. How does that sound?

jfowkes: Having looked at the `long int` floating around the Python interfaces, this is going to take a lot of effort to fix up for 64bit. Maybe we should focus on getting the interfaces to work on Windows with the current int32 library of GALAHAD and worry about moving over to the 64bit library of GALAHAD in a future major release?

amontoison: Ok for me. Let's fix the bugs with the int32 library for now and switch to int64 library in the future.

jfowkes: I've set aside tomorrow to debug on my Windows 11 desktop with MSYS2.

jfowkes:
I have been testing the GALAHAD Python interfaces on my Windows 11 desktop and get the following message:
```
Warning: Numpy built with MINGW-W64 on Windows 64 bits is experimental, and only available for
testing. You are advised not to use it for production.

CRASHES ARE TO BE EXPECTED - PLEASE REPORT THEM TO NUMPY DEVELOPERS
```
That being said, most of our GALAHAD Python packages do seem to work no problem. There are inevitably a few that crash out horribly, but that is not our problem but upstreams, and until they fix NumPy for Windows there's not much we can do.

For example, for `BSC` it determines that there are a ridiculous number of entries in the Schur complement for an input matrix with 6 entries and then inevitably crashes:
```
** python test: bsc
 S_ne: 25456912

Thread 1 received signal SIGSEGV, Segmentation fault.
```

Similarly, `CRO` crashes out because one of its indices ends up being a large negative number:
```
** python test: cro

Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff9e49ece4f in galahad_cro_double::cro_block_solve at ../src/cro/cro.F90:4015
4015                IF ( X_inorder( i ) < 0 .AND. X_inorder( j ) < 0 ) THEN
(gdb) p X_inorder
$1 = (-1, -2, -3, -4, -5, -6, -7, -8, -9, -10, -11)
(gdb) p i
$2 = 2
(gdb) p j
$3 = -1414812864
```

@amontoison in light of the above I propose that we close this PR and create a new one with just the first two commits, and use that to create a new bugfix release of GALAHAD (with the caveat that the Windows Python packages are experimental).

nimgould:
Any idea where the impossible s_ne is set in bsc. I believe that this originates in nnz_adat in bsc_form. Can gdb or whatever you are using tell us?

jfowkes:
I'm assuming `nnz_col_j`, it looks like another case of an index variable blowing up. I really don't see how these errors can occur without some sort of variable corruption of the input variables.

jfowkes:
Having read much more about Python and Windows yesterday, I have come to the conclusion that building with MSYS2 is the wrong approach for Python since the official Python and NumPy builds are compiled against the Microsoft Visual Studio UCRT runtime (not to be confused with the MSYS2 UCRT64 runtime based on top of mingw-w64).

jfowkes:
@amontoison my AI friend tells me that in order to interoperate with MSVC-ABI Python we should:

> use Intel Fortran (ifx) to build GALAHAD on Windows instead of gfortran — ifx/ifort are MSVC-ABI/CRT compatible, eliminating the whole class of cross-runtime problems

How do you feel about that and what would the implications for Julia be?

amontoison:
At the end we use `Yggdrasil` for `GALAHAD_jll.jl` and `cipythonwheels` for Python packages, I am fine to use different compilers for CI here.
We just to be careful and enable / disable the tests (for Python or Julia) depending on the compiler on Windows.

PR #548, opened 2026-08-15
Use i4_ (not ip_) for OpenMP runtime externals

amontoison:
cc @nimgould


nimgould (commented):
Not sure why we need the 
WRITE( out, "( ' number of threads = ', I0 )" ) INT( OMP_GET_MAX_THREADS( ), KIND = ip_ )
I believe that I0 takes its kind from the argument. But certainly we do need the  INTEGER ( KIND = i4_ ) :: OMP_GET_MAX_THREADS

amontoison:
I updated the PR, Nick 👍

nimgould:
Thank you. Perfect

PR #549, opened 2026-08-17
[CI] Update check_files.jl

amontoison:
@nimgould Should we still maintain `src/README.packages` and `src/README.status` ?


nimgould:
I am happy if you want to remove README.status, but I do refer to README.packages

PR #551, opened 2026-09-11
Support NumPy 2.x 64bit integers on Windows

jfowkes:
NumPy 2.x has changed the default integer type on Windows from 32bit (NPY_LONG) to 64bit (NPY_INTP):
https://numpy.org/devdocs/numpy_2_0_migration_guide.html#windows-default-integer

This PR fixes our Python interfaces to support both the old NumPy 1.x 32bit integer and the new Numpy 2.x 64bit integer.

I also suggest we do a new bugfix release of GALAHAD v5.5.2 after merging this PR so that we can update PyPI with Python 3.14 wheels (precompiled binaries) since 3.14 is the current stable release of Python and we should really be supporting it:
https://devguide.python.org/versions/


amontoison (approved):

jfowkes:
@amontoison should we be worried about the failing Poison tests?

amontoison:
Nope, it is like `expo`, we may have a flaky test for `slblt`.

nimgould:
The failure is in slblt_factor.F90:2350, which is

            !$omp task firstprivate(p, pp) private(nok, naok) default(shared) &
            !$omp      depend(inout: sync(p)) depend(in: sync(pp))

Thus just another case of omp not tidying up after itself, I believe; this is a know issue with omp that makes speculative mallocs.

jfowkes:
Thanks both, in that case are we happy to merge this and make a new bugfix release of GALAHAD?

amontoison:
It is fine for me @jfowkes.
We just added SLBLT since 5.5.1 but I am fine to do a new release.

jfowkes:
Thanks @amontoison, it is more for the Python wheels than anything else.

nimgould:
OK for me, too. I won't be able to sort out the expo issues, although I suspect it is simply that the expo C tests are using the default sparse solver, and not one we know works. Normally we have something like
``` C
strcpy(control.trb_control.trs_control.symmetric_linear_solver,"sytr ");
```
blah, blah, blah to set subproblem linear solvers, but I don't think we do in expo C tests. We will need the C equivalent of 
```Fortran
control%SSLS_control%symmetric_linear_solver = symmetric_linear_solver
control%TRU_control%TRS_control%definite_linear_solver                      &
     = definite_linear_solver
control%TRU_control%TRS_control%symmetric_linear_solver                     &
     = symmetric_linear_solver
```
where
```Fortran 
CHARACTER ( len = 5 ) :: symmetric_linear_solver = 'sytr '
CHARACTER ( len = 5 ) :: definite_linear_solver = 'potr '
```
You might try that (both in expo/C/expot.c and expotf.c)

jfowkes:
Thanks both, in that case I propose a v5.5.2 release to fixup and update the Python wheels and add SBLT.

I don't think the flaky expo test should hold this up as the problem seems to be with the test setup and this should be very easy to fix (I would do this in time for the release but unfortunately I am away at a workshop all next week).

amontoison:
@jfowkes The PyPI compilation failed for the release 5.5.2:
https://github.com/ralna/GALAHAD/actions/runs/34624008343/job/103453465709

jfowkes:
@amontoison it's a meson issue, would you be able to have a look? I'm away at a workshop all this week:
```
../meson.build:25:4: ERROR: Problem encountered: NumPy requires GCC >= 10.3
```

amontoison:
@jfowkes It is not really a Meson issue.
GCC 10.2 is installed in the wheels and Numpy requires gcc 10.3. You need to upgrade gcc in the script for the wheels.

I don't have any bandwidth for GALAHAD this week, you can work on it when you come back. :)

jfowkes:
Thanks @amontoison I've bumped the version of `cibuildhweel` and triggered a rebuild of the wheels, fingers crossed 🤞

